PR #45, opened 2021-04-01
Make a single `pytest.ini` file for all extensions.

cqc-alec:
Rather than duplicate the same file in each extension.


ss2165:
Will pytest just find it correctly?

cqc-alec:
> Will pytest just find it correctly?

Yes I just checked: it must search up the directory tree.

ss2165 (approved):

PR #46, opened 2021-04-01
[pytket-qsharp] Remove explicit call to test function.

cqc-alec:
This causes the test to be run during test _collection_, which isn't desirable.


ss2165 (approved):
my bad, sorry!

(Though PR title should say [pytket-qsharp])

Issue #32, opened 2021-03-17
[pytket-cirq] Unused qubits not included in state returned by CirqStateSimBackend


ss2165:
Can be seen in comparison with AerStateBackend: 

```python
from pytket.extensions.cirq import CirqStateSimBackend
from pytket.extensions.qiskit import AerStateBackend
from pytket import Circuit
cirq_state = CirqStateSimBackend()
c = Circuit(2).X(1)
print(cirq_state.get_state(c))
aer_state = AerStateBackend()
print(aer_state.get_state(c))
```

```
array([0.+0.j, 1.+0.j)
array([0.+0.j, 1.+0.j, 0.+0.j, 0.+0.j])
```

PR #48, opened 2021-04-08
[pytket-cirq] Update Cirq DensityMatrix/State simulators to include all circuit qubits

sjdilkes:
Previously qubits that had no operations were not included in simulated results - this behaviour has now been changed.
In the tk_to_cirq conversion, if a qubit has no operation in the pytket circuit then it would not be present in the cirq circuit. cirq circuit objects have no 'qubits' attribute, the cirq circuit all_qubits method actually iterates through all moments, creating a frozenset of all qubits it encounters in gates. Therefore, to "add a qubit" to a circuit that has no operations, an identity gate must be added to the circuit.

To do so without tampering with the tk_to_cirq conversion code, for the moments simulators a copy of the pytket circuit for simulating is taken and identity gates are added to all qubits. When the cirq circuit is produced, even gates without any other operations on will still be present in the cirq circuit simulated.

An alternative would be to change the tk_to_cirq conversion code to add identity gates to unused qubits, though this meant the tket circuit produced by converting this cirq circuit back using cirq_to_tk would not match the original circuit. I couldn't think of a more elegant situation, so modify a copied circuit in the called simulator methods seemed better...

### DIFF
--- a/modules/pytket-cirq/pytket/extensions/cirq/cirq_convert.py
+++ b/modules/pytket-cirq/pytket/extensions/cirq/cirq_convert.py
@@ -219,9 +219,13 @@ def tk_to_cirq(tkcirc: Circuit) -> cirq.circuits.Circuit:
                 "Cirq can only support registers of dimension <=2"
             )
     oplst = []
+    unused_qbs = set(tkcirc.qubits)
     for command in tkcirc:
         op = command.op
         optype = op.type
+        for qbit in command.args:
+            if qbit in unused_qbs:
+                unused_qbs.remove(qbit)
         try:
             gatetype = _ops2cirq_mapping[optype]
         except KeyError as error:
@@ -249,6 +253,9 @@ def tk_to_cirq(tkcirc: Circuit) -> cirq.circuits.Circuit:
                 cirqop = gatetype(exponent=params[0])(*qids)
         oplst.append(cirqop)
 
+    for q in unused_qbs:
+        oplst.append(cirq.ops.I(qmap[q]))
+
     try:
         coeff = cmath.exp(float(tkcirc.phase) * cmath.pi * 1j)
         if coeff != 1.0:
@@ -258,7 +265,6 @@ def tk_to_cirq(tkcirc: Circuit) -> cirq.circuits.Circuit:
             "Global phase is dependent on a symbolic parameter, so cannot adjust for "
             "phase"
         )
-
     return cirq.circuits.Circuit(*oplst)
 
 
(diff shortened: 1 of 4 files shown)


cqc-alec (changes requested):
* Could you add a short test that computes the state from the circuit as in https://github.com/CQCL/pytket-extensions/issues/32 and checks that the state vector is as expected?
* Another solution which I slightly prefer is to have an optional keyword argument such as `copy_all_qubits` to `tk_to_cirq`, which defaults to `False`. Behaviour is identical if this isn't provided, so the round-trip tests still pass, but we can set it to `True` in the two places where it's needed here, and avoid the need to duplicate code.

sjdilkes:
> * Could you add a short test that computes the state from the circuit as in #32 and checks that the state vector is as expected?
> * Another solution which I slightly prefer is to have an optional keyword argument such as `copy_all_qubits` to `tk_to_cirq`, which defaults to `False`. Behaviour is identical if this isn't provided, so the round-trip tests still pass, but we can set it to `True` in the two places where it's needed here, and avoid the need to duplicate code.


test_blank_wires should confirm this I think, but I have now extended the test_conversions test to explicitly confirm that we get the expected behaviour
I've now changed it to use the kwarg behaviour you suggested.

modules/pytket-cirq/pytket/extensions/cirq/cirq_convert.py
```diff
@@ -182,13 +182,17 @@ def cirq_to_tk(circuit: cirq.circuits.Circuit) -> Circuit:
...
+def tk_to_cirq(tkcirc: Circuit, copy_all_qubits: bool = False) -> cirq.circuits.Circuit:
     """Converts a tket :py:class:`Circuit` object to a Cirq :py:class:`Circuit`.
 
     :param tkcirc: The input tket :py:class:`Circuit`
 
     :return: The Cirq :py:class:`Circuit` corresponding to the input circuit
     """
+    if copy_all_qubits:
+        for q in tkcirc.qubits:
+            tkcirc.add_gate(OpType.noop, [q])
```
cqc-alec, line 194: This will modify the circuit passed in, which seems undesirable, so I think in this case we should take a copy. I think it should suffice to put `tkcirc = tkcirc.copy()` before the line `for q in tkcirc.qubits:` (but a test to verify that this method doesn't modify the original circuit would be good).

cqc-alec (changes requested):

modules/pytket-cirq/pytket/extensions/cirq/cirq_convert.py
sjdilkes: I've added a copy as suggested.
I've extended the test_conversions test to make sure that this copy behaviour is there (confirming that the converted circuit doesn't match the original circuit, i.e. the original is unchanged).

cqc-alec (approved):

PR #52, opened 2021-04-15
[pytket-qiskit] Stop using aqua in tests.

### LINKED ISSUE #15: [pytket-qiskit] Don't use `qiskit.aqua` in tests.
See https://github.com/Qiskit/qiskit-aqua/pull/1448#issuecomment-790702969.

"At the next release Aqua will be deprecated. Core content, such as algorithms and opflow have been moved to Terra, and domain logic to separate repos, such as qiskit-optimization."

We should use these modules instead of `aqua` in the tests, and also update the `TketBackend` docstring which discusses wrapping it in an `aqua.QuantumInstance`,

cqc-alec:
* Rewrite one test to use `PauliSumOp` instead of `WeightedPauliOperator`.
* Remove a test that uses algorithms no longer provided by qiskit (the `TketBackend` is tested elsewhere so hopefully we don't lose much by doing this).

Closes #15.

### DIFF
--- a/modules/pytket-qiskit/tests/qiskit_backend_test.py
+++ b/modules/pytket-qiskit/tests/qiskit_backend_test.py
@@ -12,10 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List
 import platform
 import numpy as np
-from pytket.backends import Backend
 from pytket.extensions.qiskit import (
     AerBackend,
     AerStateBackend,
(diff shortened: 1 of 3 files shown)


ss2165 (approved):
seems to have got more verbose!

cqc-alec:
> seems to have got more verbose!

To be fair, some of the verbosity is due to massaging the object to consist of a circuit containing one Instruction, in order to test what the test is supposed to be testing -- you can also just do `to_circuit()` instead of `to_instruction()` if you just want a circuit.

PR #56, opened 2021-04-22
[bugfix][pytket-cirq][tket-1240] Raise ValueError if n_shots not appropriate for _CirqSampleBackend

sjdilkes:
_CirqSampleBackend always requires a number of shots regardless of how results are produced.
For other backends where this is the case, the `process_circuits` method will raise a ValueError if n_shots is not set, or is set inappropriately. 
Currently _CirqSampleBackend does not do this, but Isobel has correctly pointed out that it probably should.
So now it does.


lmondada (approved):
I can get onboard with this!

PR #58, opened 2021-04-22
[pytket-aqt][bugfix] Fixed RebaseAQT

lmondada:
The rebase to the AQT gate set was wrong, for both single qubit and two qubit replacements.


lmondada:
Note: I am using the qiskit unitary simulator for the test I added for the time being. A better solution would be to use tketsim, once that is available in Python [TKET-1262]. Jira ticket: [TKET-1265]

cqc-melf (approved):

PR #68, opened 2021-04-30
Correct phase error in IBM rebase and support symbolic angles.

ss2165:
Also use fixtures in the backend_test to only construct the backend once (which is a network limited step).

### DIFF
--- a/modules/pytket-qiskit/tests/backend_test.py
+++ b/modules/pytket-qiskit/tests/backend_test.py
@@ -886,3 +886,33 @@ def test_symbolic_rebase() -> None:
     _rebase_pass.apply(pytket_circ)
 
     assert len(pytket_circ.free_symbols()) == 2
+
+
+def _tk1_to_rotations(a: float, b: float, c: float) -> Circuit:
+    """Translate tk1 to a RzRxRz so AerUnitaryBackend can simulate"""
+    circ = Circuit(1)
+    circ.Rz(c, 0).Rx(b, 0).Rz(a, 0)
+    return circ
+
+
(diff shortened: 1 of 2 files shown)


modules/pytket-qiskit/tests/backend_test.py
```diff
@@ -860,3 +868,84 @@ def test_compilation_correctness() -> None:
...
+    (7.5, 1.3, 1.5),
+    (3.5, 7.7, 0.5),
+    (8.5, 1.3, 2.5),
+]
+
+
+@pytest.mark.parametrize("a,b,c", testvals)
+def test_rebase_phase(a: float, b: float, c: float) -> None:
+    backend = AerUnitaryBackend()
+    assert _verify_single_q_rebase(backend, a, b, c)
```
cqc-alec, line 951:
```suggestion
def test_rebase_phase() -> None:
    backend = AerUnitaryBackend()
    for a in [0.6, 0, 1, 2, 3]:
        for b in [0.7, 0, 0.5, 1, 1.5]:
            for c in [0.8, 0, 1, 2, 3]:
                assert _verify_single_q_rebase(backend, a, b, c)
```
This would give me greater confidence that we've covered all bases, if this should be refactored in the future.

cqc-alec (changes requested):

ss2165:
Thank you, I've edited the commits to make the suggested changes.

To the last commit I've also added a rebase case for Rx(0.5) ~ Sx, which was somehow missed.

cqc-alec (approved):

PR #70, opened 2021-05-04
Enable contextual optimizations from backends.

cqc-alec:


modules/pytket-honeywell/tests/backend_test.py
```diff
@@ -235,6 +235,26 @@ def test_classical() -> None:
...
 
 
+@pytest.mark.skipif(skip_remote_tests, reason=REASON)
+def test_postprocess() -> None:
+    b = HoneywellBackend("HQS-LT-S1-APIVAL")
+    assert b.supports_contextual_optimisation
+    c = Circuit(2, 2)
+    c.add_gate(OpType.PhasedX, [1, 1], [0])
+    c.add_gate(OpType.PhasedX, [1, 1], [1])
+    c.add_gate(OpType.ZZMax, [0, 1])
```
ss2165, line 245: why not do the same bell test and check the outsput shots?

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -345,28 +368,45 @@ def process_circuits(
...
+                qcs.append(tk_to_qiskit(c0))
+                ppcirc_strs.append(str(ppcirc_rep))
             qobj = assemble(qcs, shots=n_shots, memory=self._config.memory)
             if self._MACHINE_DEBUG:
                 handle_list += [
-                    ResultHandle(_DEBUG_HANDLE_PREFIX + str((c.n_qubits, n_shots)), i)
+                    ResultHandle(
+                        _DEBUG_HANDLE_PREFIX + str((c.n_qubits, n_shots)),
+                        i,
+                        ppcirc_strs[i],
```
ss2165, line 399: would prefer to use `zip` over `enumerate`

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -345,28 +368,45 @@ def process_circuits(
...
+                    )
                     for i, c in enumerate(filtchunk)
                 ]
             else:
                 job = self._backend.run(qobj)
                 jobid = job.job_id()
-                handle_list += [ResultHandle(jobid, i) for i in range(len(filtchunk))]
-        for handle in handle_list:
-            self._cache[handle] = dict()
```
ss2165, line 369: does this mean this backend no longer uses `self._cache`?

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibmq_emulator.py
```diff
@@ -64,3 +83,71 @@ def required_predicates(self) -> List["Predicate"]:
 
     def default_compilation_pass(self, optimisation_level: int = 1) -> "BasePass":
         return self._ibmq.default_compilation_pass(optimisation_level)
+
+    @property
+    def _result_id_type(self) -> _ResultIdTuple:
+        return (str, int, str)
+
+    def process_circuits(
```
ss2165, line 91: why not just support postprocessing in AerBackend? Is the main reason inefficiency?

ss2165 (changes requested):

modules/pytket-aqt/pytket/extensions/aqt/backends/aqt.py
```diff
@@ -176,6 +179,9 @@ def default_compilation_pass(self, optimisation_level: int = 1) -> BasePass:
                     _aqt_rebase(),
                     RemoveRedundancies(),
                     EulerAngleReduction(OpType.Ry, OpType.Rx),
+                    SimplifyInitial(
```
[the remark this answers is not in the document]
cqc-alec: `SimplifyInitial` includes `RemoveRedundancies` by defualt, but it might introduce an `Rx` gate at the start, which could potentially be reduced by `EulerAngleReduction`. So I think replacing `RemoveRedundancies` with `SimplifyInitial` would be the best approach.

modules/pytket-braket/pytket/extensions/braket/backends/braket.py
```diff
@@ -490,24 +512,26 @@ def process_circuits(
...
             self._cache[handle] = results
             handles.append(handle)
         return handles
 
     def circuit_status(self, handle: ResultHandle) -> CircuitStatus:
         if self._device_type == _DeviceType.LOCAL:
             return CircuitStatus(StatusEnum.COMPLETED)
-        task_id, want_state = handle
+        task_id, want_state, ppcirc_str = handle
+        ppcirc_rep = literal_eval(ppcirc_str)
```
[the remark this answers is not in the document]
cqc-alec: I'm not sure there is a safety issue with `literal_eval`, but json strings would at least be a more standardized format. Will change. We could do the same for our other uses of `literal_eval` but I will leave that for now.

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -332,6 +352,8 @@ def _update_cache_result(self, handle: ResultHandle, res: BackendResult) -> None
     def circuit_status(self, handle: ResultHandle) -> CircuitStatus:
         self._check_handle_type(handle)
         jobid = str(handle[0])
+        ppcirc_rep = literal_eval(cast(str, handle[1]))
+        ppcirc = Circuit.from_dict(ppcirc_rep) if ppcirc_rep is not None else None
```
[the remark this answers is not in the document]
cqc-alec: Yes.

modules/pytket-honeywell/tests/backend_test.py
cqc-alec: I don't think this would work, as `APIVAL` doesn't actually do a simulation?

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibm.py
cqc-alec: Don't we we need enumerate to get the index `i` (second term)?

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibm.py
cqc-alec: No, `_cache` is still used as before, but now there is an additional cache. This setting of the `_cache` entries to the empty dictionary seemed unnecessary, unless I missed something.

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibmq_emulator.py
cqc-alec: Yes, mainly efficiency. The only advantage would be for AerBackends with a custom noise model. Perhaps we could consider this in the future but I think it would only really be useful if the gateset were customizable too.

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibm.py
ss2165: Ah yeah true, you can do `enumerate(zip())` but this is fine

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibm.py
ss2165: It was a record of "this job has been submitted but not completed", a record of the handles this backend has produced, but I guess we don't really use it. It is supposed to happen in all backends though.

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibmq_emulator.py
ss2165: Not sure what you mean about customisable gateset?

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibm.py
cqc-alec: OK I can put this back.

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibmq_emulator.py
cqc-alec: Well, the use case I see for an AerBackend with noise model is to simulate some actual device which will have a more limited gate set than the AerBackend. For IBM devices we have the IBMQEmulatorBackend; for others we'd need to do a custom rebase pass before running on the AerBackend, so the flow is less automatic anyway.

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -352,8 +352,7 @@ def _update_cache_result(self, handle: ResultHandle, res: BackendResult) -> None
     def circuit_status(self, handle: ResultHandle) -> CircuitStatus:
         self._check_handle_type(handle)
         jobid = str(handle[0])
-        ppcirc_rep = literal_eval(cast(str, handle[1]))
-        ppcirc = Circuit.from_dict(ppcirc_rep) if ppcirc_rep is not None else None
+        ppcirc_rep = json.loads(cast(str, handle[1]))
```
ss2165, line 355: this load only needs to happen after completion too right?

cqc-alec: Oops, yes.

ss2165 (approved):

PR #78, opened 2021-05-11
[pytket-qiskit] Remove deprecated usages of qiskit.

cqc-alec:
Make it so that all unit tests run without warnings.


modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py
```diff
@@ -354,6 +370,7 @@ class AerBackend(_AerBaseBackend):
     _supports_shots = True
     _supports_counts = True
     _supports_expectation = True
+    _expectation_allows_nonhermitian = False
```
ss2165, line 373: I guess this is not supposed to be set for every backend? maybe a comment to say this is a variable just for this backend.

modules/pytket-qiskit/pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -427,6 +440,11 @@ def append_tk_command_to_qiskit(
         qcirc.append(new_gate, qargs)
         return qcirc
 
+    if optype == OpType.CU3:
+        params = [param_to_qiskit(p, symb_map) for p in op.params] + [0]
+        params.append(0)
```
ss2165, line 445: should 0 be appended twice?

modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py
```diff
@@ -18,7 +18,6 @@
 from typing import Dict, Iterable, List, Optional, Tuple, cast, TYPE_CHECKING, Set
 
 import numpy as np
-import qiskit.providers.aer.extensions.snapshot_expectation_value  # type: ignore # pylint: disable=unused-import
```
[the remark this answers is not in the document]
cqc-alec: Right, it is replaced with `save_expectation_value` (which is also magical).

modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py
cqc-alec: Currently the default is `True`, so if non-hermitian isn't supported we have to set it to `False`. Maybe this is the wrong way round.

modules/pytket-qiskit/pytket/extensions/qiskit/qiskit_convert.py
cqc-alec: Ooh, not sure how that worked...

cqc-alec: My first thought was that it wasn't being tested, but it is, so it must be that qiskit just ignores the extra parameter.

modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py
ss2165: I thought backends by default didn't support expectation values at all?

cqc-alec: Right, but this variable is only meaningful for those that do.

ss2165: I find the following in `backend.py`
```
   _supports_expectation = False
    _expectation_allows_nonhermitian = True
```
to be confusing, if we can flip the default I think it is less likely to lead to misunderstanding.

cqc-alec: Yeah I agree. Can we make a new task to change that (since it will require changes in both tket and extensions)?

cqc-alec: Or I could make all expectation-supporting backends set this explicity in this PR, and later we can change the default?

ss2165: I think doing them together with a new task is fine.

ss2165 (approved):

PR #88, opened 2021-05-19
Using a forked sphinx 4.0.1. 

### LINKED ISSUE #84: [pytket-cirq] docs build fails
Merging #83 seems to have caused the docs build to fail for pytket-cirq (both on CI and locally).
```
.../pytket-extensions/modules/pytket-cirq/docs$ sphinx-build -b html . build
Running Sphinx v4.0.1
[autosummary] generating autosummary for: api.rst, changelog.rst, index.rst

Extension error (sphinx.ext.autosummary):
Handler <function process_generate_options at 0x7fb5485c71f0> for event 'builder-inited' threw an exception (exception: list.remove(x): x not in list)
```

yao-cqc:
This PR should fix #84.

In Sphinx autodoc v4.0.1, if a program mutate `sys.meta_path` then the `sys.meta_path.remove(finder)` in this [line](https://github.com/sphinx-doc/sphinx/blob/bdce720bb9f69d1c39ed14d214e15db4df341805/sphinx/ext/autodoc/mock.py#L151) might fail (raise a ValueError). Unfortunately, Cirq 0.11.0 does [exactly](https://github.com/quantumlib/Cirq/blob/58e16054159af6719825ada1d477b3b6d95204c5/cirq-core/cirq/_compat.py#L557) that to handle depracated modules.

This PR uses a [forked](https://github.com/yao-cqc/sphinx/blob/7691d43d0aebe1f339e4bd2704d2237295507fa1/sphinx/ext/autodoc/mock.py#L151) Sphinx v4.0.1 that ignores the exception rased by `sys.meta_path.remove(finder)`. Not removing the mock finders should only affect the speed of importing modules. (i.e. the `sys.meta_path` keeps getting larger, a _guess_ according to [this](https://docs.python.org/3/library/sys.html#sys.meta_path))

Another possible fix is to use a forked Cirq 0.11.0 with [this line](https://github.com/quantumlib/Cirq/blob/58e16054159af6719825ada1d477b3b6d95204c5/cirq-core/cirq/_compat.py#L557) removed only for building the `pytket-cirq` docs; but then the fork needs to be updated for every Cirq release.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,2 @@
-sphinx
+git+git://github.com/yao-cqc/sphinx.git@ignore-mutated-finders-in-autodoc
 sphinx_rtd_theme


cqc-alec:
I was curious to see how google got around this problem and I saw this recent commit:
https://github.com/quantumlib/Cirq/commit/fd257e6c228a0b4c8839ad6ac3d30467476f195e#diff-221cc76e1b4126c2205cba4d26c274d8cac8015525502d451923f9cf91ea2fd5R186
I don't know what this is doing but maybe it is a trick we can copy?
Anyway, let's go with this as the short-term fix to get the build working, Thank you!

cqc-alec (approved):

PR #93, opened 2021-05-24
[pytket-braket] Allow BraketBackend to take an AwsSession

### LINKED ISSUE #90: pytket-braket relies on setting credentials in the AWS CLI
I'd like to use pytket-braket in an automated fashion, and so it would be good to have a way to pass credentials into `BraketBackend` programmatically, rather than relying on the AWS CLI.

This might be doable if `BraketBackend` took an optional argument of a [boto3 Session](https://github.com/boto/boto3/blob/27da6f36e6afcb3453b4a6ae8c6813b490e4bf55/boto3/session.py#L28), which can take credentials directly, or a [braket AwsSession](https://github.com/aws/amazon-braket-sdk-python/blob/dde5df0a7baa0f5df118cd865ab622fcca1da1cb/src/braket/aws/aws_session.py#L25) which can take a boto3.Session.

I'll work on a PR for this if that's okay.

isobelhooper:
Resolves #90.

BraketBackend can now take a `braket.aws.AwsSession` object, which itself can take a `boto3.Session` object. Since we [can pass AWS credentials to the `boto3.Session` object](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#passing-credentials-as-parameters), this means we can use BraketBackend without having credentials set as environment variables (as the AWS CLI does).

If the user doesn't pass an AwsSession object, BraketBackend behaves as it did before (with braket's internal code creating sessions as needed).

----

I've tested this using the Amazon Braket SV1 simulator in us-west-1 (without any other AWS credentials configured) and it seems to work okay. I'm not sure that it would help much to add a unit test in for this functionality as the remote tests aren't run by default, but it'd be relatively easy to test manually.

### DIFF
--- a/modules/pytket-braket/pytket/extensions/braket/backends/braket.py
+++ b/modules/pytket-braket/pytket/extensions/braket/backends/braket.py
@@ -62,7 +62,7 @@ from pytket.utils import prepare_circuit
 from pytket.utils.operators import QubitPauliOperator
 from pytket.utils.outcomearray import OutcomeArray
 import braket  # type: ignore
-from braket.aws import AwsDevice  # type: ignore
+from braket.aws import AwsDevice, AwsSession  # type: ignore
 from braket.aws.aws_device import AwsDeviceType  # type: ignore
 from braket.aws.aws_quantum_task import AwsQuantumTask  # type: ignore
 import braket.circuits  # type: ignore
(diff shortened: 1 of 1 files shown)


modules/pytket-braket/pytket/extensions/braket/backends/braket.py
```diff
@@ -219,6 +220,7 @@ def __init__(
             default: "quantum-simulator"
         :param provider: provider name from device ARN (e.g. "ionq", "rigetti", ...),
             default: "amazon"
+        :param aws_session: braket AwsSession object, to pass credentials in
```
cqc-alec, line 223: Can we add something like "if not configured on local machine", so as not to give the impression people need to use this?

cqc-alec (changes requested):

cqc-alec (approved):

PR #95, opened 2021-05-25
Update pytket version requirements, versions and changelogs.

cqc-alec:


modules/pytket-cirq/docs/changelog.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 ~~~~~~~~~
 
+0.12.0 (May 2021)
+-----------------
+
+* Updated pytket version requirement to 0.11.
+
 0.11.0 (unreleased)
```
cqc-melf, line 9: I this correct?

cqc-melf: I was wondering about the "unreleased"

cqc-alec: Yes, that version was not released.

cqc-alec:
I think the Linux build failure is because:
* It installed and tested pytket-qulacs before pytket-cirq.
* The pytket-qulacs tests have a dependency on openfermion.
* openfermion has a dependency on cirq 0.10.
* So when cirq 0.11 gets installed it is replacing an installation of cirq 0.10.
* Which causes horrible things to happen on Linux, as we have already found.

cqc-alec:
The fact that openfermion (needed by the qulacs tests) need an older version of cirq, which then has to be overwritten, is causing problems (again). Need to find a way to work around this.

cqc-alec:
The last commit just ensures that we test pytket-cirq before pytket-qulacs. This works around the problem, but it would be good to find a better long-term solution.

cqc-melf (approved):

PR #103, opened 2021-06-07
[pytket-braket] Add @property decorator to BraketBackend.device

### LINKED ISSUE #102: [pytket-braket] BraketBackend.device is missing a @property decorator
This means that calling it works differently to all the other backends in this repository, which do have `@property` for their `.device` method.

Will submit a PR to address this.

isobelhooper:
Fixes #102. 

This matches the definition of `.device` in other Backend classes.


cqc-alec (dismissed):
Thanks!

cqc-alec:
Just need to fix the call to `device()` in `test_ionq`.

cqc-alec (approved):

PR #119, opened 2021-06-18
Install types modules.

cqc-alec:


Roland-djee:
Weird, I can't add my review.

cqc-alec:
> Weird, I can't add my review.

Sorry, you were automatically added as a codeowner along with several others; I removed these and left @isobelhooper  as this affects her PR.

modules/pytket-honeywell/setup.py
```diff
@@ -40,9 +40,11 @@
     install_requires=[
         "pytket ~= 0.11.0",
         "requests >= 2.2",
+        "types-requests",
         "websockets >= 7.0",
         "nest_asyncio >= 1.2",
         "pyjwt >= 1.7, < 2.0",
+        "types-jwt",
```
isobelhooper, line 47: Looks good, although we'll need to remember to remove this should we use pyjwt 2.0+, according to https://pypi.org/project/types-jwt/.

isobelhooper (approved):
Changes look good!

modules/pytket-honeywell/setup.py
cqc-alec: Yes, good point.

PR #139, opened 2021-07-14
Update use of BackendInfo to reflect pytket/Mushroom requests

sjdilkes:
Or explicitly, update pytket-qiskit, pytket-braket and pytket-pyquil to use new BackednInfo attributes instead of the characterisation bucket dictionary.


yao-cqc (approved):
Looks good! Thanks

PR #146, opened 2021-07-19
[all] Update extensions to pytket 0.13.0rc1

ss2165:


modules/pytket-honeywell/tests/backend_test.py
```diff
@@ -51,7 +51,7 @@
 
 def test_honeywell() -> None:
     backend = HoneywellBackend(
-        device_name="HQS-LT-1.0-APIVAL", machine_debug=skip_remote_tests
+        device_name="HQS-LT-S1-APIVAL", machine_debug=skip_remote_tests
```
cqc-alec, line 54: What is the difference?

ss2165: the previous is an older device, which is no longer available for testing

cqc-alec: ok

cqc-alec (approved):

PR #167, opened 2021-08-13
support new honeywell simulator options

ss2165:
Simulator type (state-vector or stabilizer) specified on backend
initialisation, to perhaps support compilation and predicate checking in
future.
Whether to run noisy simulation specified at process_circuit time via
kwarg.

### DIFF
--- a/modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
+++ b/modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
@@ -57,11 +57,7 @@ from .config import set_honeywell_config
 from .api_wrappers import HQSAPIError, HoneywellQAPI
 
 _DEBUG_HANDLE_PREFIX = "_MACHINE_DEBUG_"
-
-
 HONEYWELL_URL_PREFIX = "https://qapi.honeywell.com/v1/"
-
-HONEYWELL_DEVICE_APIVAL = "HQS-LT-1.0-APIVAL"
 
 _STATUS_MAP = {
     "queued": StatusEnum.QUEUED,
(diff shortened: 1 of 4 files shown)


modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -88,6 +88,10 @@
 }
 
 
+class GetResultFailed(Exception):
```
cqc-melf, line 91: Is this the intended function?

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -61,7 +61,7 @@
 
 HONEYWELL_URL_PREFIX = "https://qapi.honeywell.com/v1/"
 
-HONEYWELL_DEVICE_APIVAL = "HQS-LT-1.0-APIVAL"
+HONEYWELL_DEVICE_APIVAL = "HQS-LT-S1-APIVAL"
```
cqc-melf, line 64: This variable is unused. Is that intended ?

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
ss2165: yes

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
ss2165: no, will remove

cqc-melf (approved):

PR #183, opened 2021-09-09
Update pytket-cirq to support cirq 0.12

sjdilkes:
As the title says, includes the inclusion of cirq noise models for density matrix simulator backends.

### DIFF
--- a/modules/pytket-cirq/pytket/extensions/cirq/backends/cirq.py
+++ b/modules/pytket-cirq/pytket/extensions/cirq/backends/cirq.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 from abc import abstractmethod
-from typing import Sequence, cast, Optional, List, Sequence, Union
+from typing import Sequence, cast, Optional, List, Union
+from numpy import complex64
 from uuid import uuid4
 from cirq.sim import (
     CliffordSimulator,
@@ -25,7 +26,6 @@ from cirq.sim import (
     StateVectorTrialResult,
 )
 
-# import cirq
 from cirq import ops
 from cirq.circuits import Circuit as CirqCircuit
 
@@ -162,7 +162,6 @@ class _CirqSampleBackend(_CirqBaseBackend):
             optional=False,
         )
 
-        self._simulator = self.set_simulator(seed=kwargs.get("seed"))
         if valid_check:
             self._check_all_circuits(circuits)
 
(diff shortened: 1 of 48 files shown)


modules/pytket-cirq/pytket/extensions/cirq/backends/cirq.py
```diff
@@ -54,9 +53,9 @@
 from pytket.backends.backendinfo import BackendInfo
 from pytket.backends.resulthandle import _ResultIdTuple
 from pytket.utils.results import KwargTypes
-from pytket.extensions.cirq import tk_to_cirq
+from pytket.extensions.cirq.cirq_convert import tk_to_cirq  # type: ignore
 from pytket.utils.outcomearray import OutcomeArray
-from .cirq_utils import _get_default_uids
+from .cirq_utils import _get_default_uids  # type: ignore
```
lmondada, line 58: This shouldn't need a type ignore. Do you know why it does? (especially since it seems it wasn't necessary before)

modules/pytket-cirq/pytket/extensions/cirq/backends/cirq.py
```diff
@@ -370,7 +355,8 @@ def package_results(
         )
         all_backres = [
             BackendResult(
-                state=cast(SparseSimulatorStep, run).state_vector(copy=True),
+                # state=cast(SparseSimulatorStep, run).state_vector(copy=True),
```
lmondada, line 358: I think you can remove the comment, or is there a reason for leaving it?

modules/pytket-cirq/pytket/extensions/cirq/cirq_convert.py
```diff
@@ -22,7 +22,7 @@
 from logging import warning
 from cirq.devices import LineQubit, GridQubit
 import cirq.ops
-import cirq_google
+import cirq_google  # type: ignore
```
lmondada, line 25: Here as well, I'm just confused why the type checks had to be changed. Is it that Cirq has newly introduced type annotations?

modules/pytket-cirq/tests/test_cirq_backend.py
```diff
@@ -17,19 +17,18 @@
 import math
 from hypothesis import given, strategies
 import numpy as np
-from pytket.extensions.cirq import (
+from pytket.extensions.cirq.backends.cirq import (  # type: ignore
```
lmondada, line 20: same here, sorry for bugging you on this. I'm sure there is a one-line answer to my annoyingness, at least.

lmondada (changes requested):
Honestly, all good, except I'm being a prick about mypy type checks. I should relent if you provide me with some answers :)

modules/pytket-cirq/pytket/extensions/cirq/backends/cirq.py
sjdilkes: you were right, think my local build is turbo cursed

lmondada (approved):
sorry for the delay!

PR #185, opened 2021-09-09
allow expectation values to be evaluated for QubitPauliOperator with …

irfankhan10:
Allow `Backend.get_expectation_value(self, operator: QubitPauliOperator)` to work for `QubitPauliOperator` with either  `Dict[QubitPauliString, float]` or `Dict[QubitPauliString, sympy.Expr]`.


irfankhan10:
@ss2165 @cqc-alec pls review as well. 

We need this because in our code we have child of `QubitPauliOperator` that sometimes returns Dict[QubitPauliString, float]  instead of Dict[QubitPauliString, Expr].

This means the `get_operator_expectation_value` raises an `AttributeError`, since python `floats` do not have` .evalf()`

cqc-alec (approved):

Issue #180, opened 2021-09-07
[pytket-qsharp] import qsharp fails


cqc-melf:
On windows with python 3.8 the import of qsharp fails with the following error message:

from qsharp import compile as qscompile  # type: ignore

C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\site-packages\qsharp_init_.py:198: in
client = start_client()
C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\site-packages\qsharp\clients_init.py:34: in _start_client
client = qsharp.clients.iqsharp.IQSharpClient(kernel_name=kernel_name)
E AttributeError: partially initialized module 'qsharp' has no attribute 'clients' (most likely due to a circular import)

The fail occurs on the github ci:

We tried different qsharp version, for example 0.18.2107.153439 and 0.18.2108.160999. The error occurs in all of them.

https://github.com/CQCL/pytket-extensions/runs/3534918995?check_suite_focus=true

The problem was reported to qsharp on:
https://github.com/microsoft/iqsharp/issues/512

cqc-alec:
Still an issue with qsharp 0.20.

PR #204, opened 2021-09-30
[pytket-qsharp] Use `actions/setup-dotnet` in workflows.

### LINKED ISSUE #188: [pytket-qsharp] Use `actions/setup-dotnet` in CI workflows.
Instead of installing dotnet SDK in different ways on different platforms, we can just use:
```
    - name: Install dotnet SDK
      uses: actions/setup-dotnet@v1
      with:
        dotnet-version: '3.1.x'
```
in the workflow scripts.

cqc-alec:
Also enable testing of pytket-qsharp on Python 3.7, since that is the one configuration not affected by #180 .
Closes #188 .
Tested: https://github.com/CQCL/pytket-extensions/actions/runs/1290980057 and https://github.com/CQCL/pytket-extensions/actions/runs/1291025849

### DIFF
--- a/.github/workflows/build-test
+++ b/.github/workflows/build-test
@@ -52,7 +52,8 @@ for MODULE in $MODULES_TO_TEST; do
     # FIXME see
     # https://github.com/CQCL/pytket-extensions/issues/180 and 
     # https://github.com/microsoft/iqsharp/issues/512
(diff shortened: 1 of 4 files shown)


yao-cqc (approved):
Looks good

PR #218, opened 2021-10-13
Upgrade qiskit to 0.31.0

### LINKED ISSUE #203: [pytket-qiskit] Update to qiskit 0.30.

RolandMacDoland:
Fixes #203.

This PR addresses the upgrade of the `Qiskit` dependency to version 0.31.0 in the `pytket-qiksit` extension. 
All test pass locally.
I also added a `base-test-requirements.txt` that contain common dependencies for testing extensions. I can harmonise across all extensions and update the `README.md` if found useful.

### DIFF
--- a/modules/pytket-projectq/pytket/extensions/projectq/backends/projectq_backend.py
+++ b/modules/pytket-projectq/pytket/extensions/projectq/backends/projectq_backend.py
@@ -91,6 +91,7 @@ _GATE_SET = {
     OpType.Ry,
     OpType.Rz,
     OpType.Barrier,
+    OpType.Measure,
 }
 
 


ss2165 (dismissed):

cqc-alec:
Can you add an entry in the changelog please?

modules/pytket-qiskit/tests/test-requirements.txt
```diff
@@ -1 +1,2 @@
-pytest-timeout ~= 1.4.2 
\ No newline at end of file
+-r ../../../base-test-requirements.txt
```
cqc-alec, line 1: Can we make this a separate PR, changing all extensions at once rather than just this one?
Currently these are installed in the `build-test` script; if we do make this change then we should remove the installation from there.

cqc-alec: I would prefer this change to be in a separate PR though.

PR #220, opened 2021-10-14
Feature/query available devices from qiskit

yao-cqc:


modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -340,6 +307,57 @@ def characterisation(self) -> Dict[str, Any]:
...
+        )
+        return backend_info
+
+    @classmethod
+    def available_devices(cls, **kwargs: KwargTypes) -> List[BackendInfo]:
+        hub = kwargs.get("hub")
+        group = kwargs.get("group")
+        project = kwargs.get("project")
+        IBMQ.load_account()
+        provider = IBMQ.get_provider(hub=hub, group=group, project=project)
```
ss2165, line 355: I suggest turning _get_provider in to a static function which takes an optional QiskitConfig instance, and use that here
this method should take an optional provider (like __init__) and use that if provided.

PR #224, opened 2021-10-18
[pytket-pyquil] Modify `ForestBackend` constructor to accept a `QuantumComputer`.

cqc-alec:


modules/pytket-pyquil/pytket/extensions/pyquil/backends/forest.py
```diff
@@ -100,19 +100,15 @@ class ForestBackend(Backend):
     _persistent_handles = True
     _GATE_SET = {OpType.CZ, OpType.Rx, OpType.Rz, OpType.Measure, OpType.Barrier}
 
-    def __init__(self, qc_name: str, simulator: bool = True):
-        """Backend for running circuits on a Rigetti QCS device or simulating with the
-        QVM.
+    def __init__(self, qc_name: str):
```
cqc-alec, line 103: On further reflection I think this should be constructed from the `QuantumComputer` object; this allows the possibility of accessing QPUs via the extension, even if we can't test it.

PR #239, opened 2021-11-10
Feature/implement available_devices braket

yao-cqc:


cqc-alec (changes requested):

modules/pytket-braket/pytket/extensions/braket/backends/braket.py
```diff
@@ -668,6 +669,55 @@ def characterisation(self) -> Optional[Dict[str, Any]]:
     def backend_info(self) -> BackendInfo:
         return self._backend_info
 
+    @classmethod
+    def available_devices(cls, **kwargs: Any) -> List[BackendInfo]:
+        """
+        See :py:meth:`pytket.backends.Backend.available_devices`.
+        Supported kwargs: `region` (default none).
```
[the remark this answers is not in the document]
cqc-alec: > When constructing an `AwsSession` a [client](https://github.com/aws/amazon-braket-sdk-python/blob/35e3eade19b83c309590b0f967c00e3bfcbefc32/src/braket/aws/aws_session.py#L41) is initialized (if not provided as an argument). This client is later used to fetch all devices. `Braket` docs says that every Client is associated with a single region. If a region is not specified, the `botocore` will try to [resolve](https://github.com/boto/botocore/blob/2a7d0af334c8ebadc5beaf1f44fe2aaf8256749b/botocore/session.py#L804) a region name. (possibly from the aws config file).
> 
> I think another option is to iterate through all available regions (obtained from `boto3).

Up to you -- either a comment explaining the above or iterating through them. It's not obvious to me which is more useful.

modules/pytket-braket/pytket/extensions/braket/backends/braket.py
```diff
@@ -625,6 +670,58 @@ def characterisation(self) -> Optional[Dict[str, Any]]:
     def backend_info(self) -> BackendInfo:
         return self._backend_info
 
+    @classmethod
+    def available_devices(cls, **kwargs: Any) -> List[BackendInfo]:
+        """
+        See :py:meth:`pytket.backends.Backend.available_devices`.
+        Supported kwargs: none.
+        """
+        braket_regions = ["us-west-1", "us-west-2", "us-east-1"]
```
yao-cqc, line 679: These have to be hardcoded because there isn't an obvious way to fetch available regions for the `braket` service. The `get_available_regions` method in `boto3` returns an empty list for `braket`. I assume these regions should be quite stable.

cqc-alec: OK, thanks. I would revert to how it was before -- feels weird to hard-code them, even if they are _likely_ to be stable -- not sure if this would work in all situations, e.g. using from China? Let's leave the choice of region up to the user if they want to specify it.

yao-cqc: Yeah, I agree

modules/pytket-braket/tests/backend_test.py
```diff
@@ -377,3 +377,6 @@ def test_postprocess_ionq() -> None:
 def test_retrieve_available_devices() -> None:
     backend_infos = BraketBackend.available_devices()
     assert len(backend_infos) > 0
+    # Test annealers are filtered out.
```
cqc-alec, line 380: Does this test for that?

yao-cqc: Currently, all D-wave devices are hosted in us-west-2. This test should cover it as long as AWS doesn't change the hosting arrangement or "us-west-2" remains a valid braket region. https://docs.aws.amazon.com/braket/latest/developerguide/braket-devices.html

cqc-alec: But how are we testing that they are filtered out?

yao-cqc: The anealers don't have the `DeviceActionType.JAQCD` action, so we won't be able to get the `supportedOperations` from these devices to construct `BackendInfo`s. If the filter isn't working this test would throw an exception. But I suppose I can improve the test to further make sure they are not in the returned list.

cqc-alec: OK I see, that's fine.

cqc-alec (approved):

PR #249, opened 2021-12-07
[feature] [pytket-honeywell] use batching in process_circuits

ss2165:
~waiting for Honeywell to enable for our accounts~

### DIFF
--- a/modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
+++ b/modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
@@ -55,6 +55,7 @@ from pytket.utils.outcomearray import OutcomeArray
 
 from .config import set_honeywell_config
 from .api_wrappers import HQSAPIError, HoneywellQAPI
+from .credential_storage import PersistentStorage
 
 _DEBUG_HANDLE_PREFIX = "_MACHINE_DEBUG_"
 HONEYWELL_URL_PREFIX = "https://qapi.honeywell.com/v1/"
@@ -532,14 +533,14 @@ class HoneywellBackend(Backend):
         """
         set_honeywell_config(user_name)
 
-        keyring.set_password(HoneywellQAPI.KEYRING_NAME, user_name, pwd)  # type: ignore
+        keyring.set_password(PersistentStorage.KEYRING_SERVICE, user_name, pwd)  # type: ignore
 
     @staticmethod
     def clear_saved_login(user_name: str) -> None:
         """Delete saved password for user_name if it exists"""
         set_honeywell_config(None)
         try:
-            keyring.delete_password(HoneywellQAPI.KEYRING_NAME, user_name)  # type: ignore
+            keyring.delete_password(PersistentStorage.KEYRING_SERVICE, user_name)  # type: ignore
         except keyring.errors.PasswordDeleteError:
             raise ValueError(f"No credentials stored for {user_name}")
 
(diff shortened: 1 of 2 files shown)


modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -346,6 +350,11 @@ def process_circuits(
             body["name"] = circ.name if circ.name else f"{self._label}_{i}"
             body["program"] = honeywell_circ
             body["count"] = n_shots
+            if final_index > 0:
+                body["batch-exec"] = batch_exec
+                if i == final_index:
+                    body["batch-end"] = True
+            print(body)
```
isobelhooper, line 357: This looks like a debugging statement that was left in - do we actually want to output `body`?

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -532,14 +545,14 @@ def save_login(user_name: str, pwd: str) -> None:
...
-        keyring.set_password(HoneywellQAPI.KEYRING_NAME, user_name, pwd)  # type: ignore
+        keyring.set_password(PersistentStorage.KEYRING_SERVICE, user_name, pwd)  # type: ignore
 
     @staticmethod
     def clear_saved_login(user_name: str) -> None:
         """Delete saved password for user_name if it exists"""
         set_honeywell_config(None)
         try:
-            keyring.delete_password(HoneywellQAPI.KEYRING_NAME, user_name)  # type: ignore
+            keyring.delete_password(PersistentStorage.KEYRING_SERVICE, user_name)  # type: ignore
```
isobelhooper, line 555: These are useful changes but they don't appear to be related to batching. Could these be in a separate PR (or a separate commit if they make sense to be grouped with the batching changes)?

isobelhooper: Separate commit makes sense - I'm happy with this change.

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -336,6 +336,9 @@ def process_circuits(
             basebody["group"] = group
 
         handle_list = []
+        # set some default max-batch-cost
+        batch_exec: Union[int, str] = 1
```
isobelhooper, line 340: Checking how the `batch-exec` field works, it seems from this code that:
* In the first circuit in a batch, `batch-exec` is set to `1`.
* In the second and subsequent circuits in a batch, `batch-exec` is set to the job ID of the first circuit in the batch, as retrieved from Honeywell's API.

Is that correct?

If so, it might be worth adding an explanatory comment here.

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -336,6 +336,9 @@ def process_circuits(
             basebody["group"] = group
 
         handle_list = []
+        # set some default max-batch-cost
```
isobelhooper, line 339: The variable definitions just after this don't seem to relate to max-batch-cost - is this a TODO comment?

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
ss2165: Yes that is correct, I've added some comments and documentation, thanks for the catch.

isobelhooper (approved):
With the extra comments and docstrings (and clarification on `max-batch-cost`), I think this looks good to go.

isobelhooper (approved):
Changes look the same after rebasing onto `develop` - good to go.

PR #255, opened 2022-01-04
Avoid shadowing variable name in loop.

cqc-alec:
We were getting incorrect module names at the top of the sidebar on the docs pages. This fixes that.


yao-cqc (approved):
good catch, thanks!

PR #260, opened 2022-01-07
[feature] [pytket-honeywell] support use of zzphase

ss2165:
~HQS API NOT ACCEPTING YET~
Requires release of pytket containing changes from https://github.com/CQCL/tket/pull/193
not all H-system backends support it so requires some gate set parametrisation
compiler options also need to be set in request body

### DIFF
--- a/modules/pytket-honeywell/tests/backend_test.py
+++ b/modules/pytket-honeywell/tests/backend_test.py
@@ -374,3 +374,33 @@ def test_submission_with_group() -> None:
     shots = b.run_circuit(c, n_shots=n_shots, group="test").get_shots()  # type: ignore
     print(shots)
     assert all(q[0] == q[1] for q in shots)
+
+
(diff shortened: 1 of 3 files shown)


modules/pytket-honeywell/tests/backend_test.py
```diff
@@ -374,3 +374,32 @@ def test_submission_with_group() -> None:
     shots = b.run_circuit(c, n_shots=n_shots, group="test").get_shots()  # type: ignore
     print(shots)
     assert all(q[0] == q[1] for q in shots)
+
+
+def test_zzphase() -> None:
```
cqc-melf, line 379: Should this test be run every time in the ci?
The backend looks like the real device to me?
Maybe add  `@pytest.mark.skipif(skip_remote_tests, reason=REASON)`

ss2165: good spot, thanks

cqc-melf (approved):

PR #265, opened 2022-01-11
Add IQM extension

cqc-alec:
Tested only on the "demo" device (which is not a simulator but just produces pseudo-random bits). Other devices may be specified by passing parameters to the `IQMBackend` constructor. Credentials (`username` and `api_key`) may be stored in the pytket config file (with the key `iqm`), or passed directly to the constructor.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,6 +3,7 @@
 /modules/pytket-cirq/ @sjdilkes
 /modules/pytket-honeywell/ @ss2165
 /modules/pytket-ionq/ @Roland-djee
+/modules/pytket-iqm/ @cqc-alec
 /modules/pytket-projectq/ @lmondada
 /modules/pytket-pyquil/ @sjdilkes @lmondada
 /modules/pytket-pyzx/ @cqc-alec
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -147,12 +147,12 @@ jobs:
         ./.github/workflows/build-test mypy
 
     - name: Set up Python 3.9
-      if: github.event_name == 'push'
+      if: github.event_name == 'push' || github.event_name == 'pull_request'
       uses: actions/setup-python@v2
       with:
         python-version: 3.9
     - name: Build and test (3.9)
-      if: github.event_name == 'push'
+      if: github.event_name == 'push' || github.event_name == 'pull_request'
       run: |
         ./.github/workflows/build-test nomypy
 
@@ -211,12 +211,12 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'push'
+      if: github.event_name == 'push' || github.event_name == 'pull_request'
       uses: actions/setup-python@v2
       with:
         python-version: 3.9
     - name: Build and test (3.9)
-      if: github.event_name == 'push'
+      if: github.event_name == 'push' || github.event_name == 'pull_request'
       shell: bash
       run: |
         ./.github/workflows/build-test nomypy
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -74,12 +74,11 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'push' || github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel')
       uses: actions/setup-python@v2
       with:
         python-version: 3.9
     - name: Build and test (3.9)
-      if: github.event_name == 'push'
+      if: github.event_name == 'push' || github.event_name == 'pull_request'
       run: |
         ./.github/workflows/build-test nomypy
     - name: Build, test and save (3.9)
--- a/modules/pytket-iqm/README.md
+++ b/modules/pytket-iqm/README.md
@@ -1,7 +1,7 @@
 # pytket-iqm
 
-[Pytket](https://cqcl.github.io/pytket) is a Python module for interfacing
-with CQC tket, a set of quantum programming tools.
+[Pytket](https://cqcl.github.io/pytket) is a Python module providing an
+extensive set of tools for compiling and executing quantum circuits.
 
 `pytket-iqm` is an extension to `pytket` that allows `pytket` circuits to be
 executed on [IQM](https://meetiqm.com/)'s quantum devices and simulators.
--- a/modules/pytket-iqm/README.md
+++ b/modules/pytket-iqm/README.md
@@ -1,0 +1,14 @@
+# pytket-iqm
+
+[Pytket](https://cqcl.github.io/pytket) is a Python module for interfacing
+with CQC tket, a set of quantum programming tools.
+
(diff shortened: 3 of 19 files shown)


modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
```diff
@@ -0,0 +1,344 @@
...
+from pytket.utils import prepare_circuit
+from pytket.utils.outcomearray import OutcomeArray
+from .config import IQMConfig
+
+_GATE_SET = {OpType.PhasedX, OpType.CZ, OpType.Measure}
+
+_DEFAULT_SETTINGS = Path(__file__).resolve().parent / "demo_settings.json"
+
+# https://iqm-finland.github.io/cirq-on-iqm/api/cirq_iqm.devices.adonis.Adonis.html
+_DEFAULT_COUPLING = [
```
cqc-melf, line 68: Maybe add this as a comment:
      QB1
       |
QB2 - QB3 - QB4
       |
      QB5

modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
```diff
@@ -0,0 +1,344 @@
...
+        ]
+
+    def default_compilation_pass(self, optimisation_level: int = 1) -> BasePass:
+        assert optimisation_level in range(3)
+        passes = [DecomposeBoxes(), FlattenRegisters()]
+        if optimisation_level == 1:
+            passes.append(SynthesiseTket())
+        elif optimisation_level == 2:
+            passes.append(FullPeepholeOptimise())
+        passes.append(_iqm_rebase())  # to satisfy MaxTwoQubitGatesPredicate
```
cqc-melf, line 173: I think this only needs to be done for optimisation_level 0?

modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
```diff
@@ -0,0 +1,344 @@
...
+        return SequencePass(passes)
+
+    @property
+    def _result_id_type(self) -> _ResultIdTuple:
+        return (bytes, str)
+
+    def process_circuits(
+        self,
+        circuits: Sequence[Circuit],
+        n_shots: Union[None, int, Sequence[Optional[int]]] = None,
```
cqc-melf, line 193: Are you sure this can handle n_shots = None?

modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
```diff
@@ -0,0 +1,344 @@
...
+            assert status is RunStatus.FAILED
+            return CircuitStatus(StatusEnum.ERROR, run_result.message)
+
+    def get_result(self, handle: ResultHandle, **kwargs: KwargTypes) -> BackendResult:
+        """
+        See :py:meth:`pytket.backends.Backend.get_result`.
+        Supported kwargs: `timeout` (default 900).
+        """
+        try:
+            return super().get_result(handle)
```
cqc-melf, line 275: I am not sure that I have full understood why this is not handle as the return in line 283?

modules/pytket-iqm/setup.py
```diff
@@ -0,0 +1,53 @@
...
+    author_email="alec.edgington@cambridgequantum.com",
+    python_requires=">=3.9",
+    url="https://github.com/CQCL/pytket-extensions",
+    description="Extension for pytket, providing access to IQM backends",
+    long_description=open("README.md").read(),
+    long_description_content_type="text/markdown",
+    license="Apache 2",
+    packages=find_namespace_packages(include=["pytket.*"]),
+    include_package_data=True,
+    install_requires=["pytket ~= 0.18.0", "iqm-client ~= 1.6"],
```
cqc-melf, line 40: Are you sure you want to use 1.7 here if available? maybe `==1.6` would be the better choice? They release a new version every month. (https://pypi.org/project/iqm-client/#history)

cqc-melf: ( `~=1.6.0` would maybe work too, not sure.)

modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
cqc-melf: GitHub broke this diagram somehow, please take a look on the linked side to see the diagram I mean.

.github/workflows/build_and_test.yml
```diff
@@ -74,12 +74,11 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'push' || github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel')
```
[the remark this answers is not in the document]
cqc-alec: This just removes a condition so it is always run.

.github/CODEOWNERS
```diff
@@ -3,6 +3,7 @@
 /modules/pytket-cirq/ @sjdilkes
 /modules/pytket-honeywell/ @ss2165
 /modules/pytket-ionq/ @Roland-djee
+/modules/pytket-iqm/ @cqc-alec
```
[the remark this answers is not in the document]
cqc-alec: Makes sense to have me there for now as I wrote it, but happy for anyone else to take over or review PRs.

modules/pytket-iqm/README.md
```diff
@@ -0,0 +1,14 @@
+# pytket-iqm
+
+[Pytket](https://cqcl.github.io/pytket) is a Python module for interfacing
+with CQC tket, a set of quantum programming tools.
```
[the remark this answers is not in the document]
cqc-alec: I may just remove "CQC".

modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
cqc-alec: OK.

modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
cqc-alec: Possibly, I will check.

modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
cqc-alec: I'll add a test to make sure.

modules/pytket-iqm/pytket/extensions/iqm/backends/iqm.py
cqc-alec: This just checks if it is in the cache, returns it if so, and otherwise raises `CircuitNotRunError`. It's a bit of a funny pattern, I agree, but seems to be followed elsewhere, e.g. in the `aqt` extension.

modules/pytket-iqm/setup.py
cqc-alec: If they are following semantic versioning conventions then 1.7 should be API-compatible with 1.6. (If not we can revisit this, but probably better to use the latest 1.x version available.)

cqc-melf:
Thank you for all the comments and the updates, from my side this is fine now!

cqc-melf (approved):

PR #266, opened 2022-01-14
Support Python 3.10 and drop 3.7 for pytket extensions.

cqc-alec:
* Update pytket to dev release including 3.10 support.
* Update workflows to test against {3.8, 3.9, 3.10} instead of {3.7, 3.8, 3.9}.
* Exclude from the 3.10 CI builds some extensions that can't yet support 3.10 because of dependencies (namely, pytket-qiskit and pytket-qsharp on Windows and pytket-qulacs on MacOS and Windows). (I will make issues to track these missing configurations so that we can add them when possible.)
* Update documentation.


.github/workflows/build-test
```diff
@@ -33,9 +33,57 @@ rm -rf ${ARTIFACTSDIR} && mkdir ${ARTIFACTSDIR}
...
+        pytket-qiskit)
+            # Cannot install qiskit-aer on Windows with 3.10.
+            [[ "${PLAT}" == "Windows" && "${PYVER}" == "3.10" ]] && return
+            ;;
+        pytket-qsharp)
+            # Cannot install qsharp on Windows with 3.10.
+            [[ "${PLAT}" == "Windows" && ${PYVER} == "3.10" ]] && return
+            ;;
+        pytket-qulacs)
+            # Cannot install qulacs on MacOS or Windows with 3.10.
```
cqc-melf, line 51: Is this documented in an issue somewhere?

.github/workflows/build-test
```diff
@@ -33,9 +33,57 @@ rm -rf ${ARTIFACTSDIR} && mkdir ${ARTIFACTSDIR}
...
 
+function unsupported() {
+    case "$1" in
+        pytket-iqm)
+            # The iqm-client package is only available for 3.9.
+            [[ "${PYVER}" != "3.9" ]] && return
+            ;;
+        pytket-qiskit)
+            # Cannot install qiskit-aer on Windows with 3.10.
+            [[ "${PLAT}" == "Windows" && "${PYVER}" == "3.10" ]] && return
```
cqc-melf, line 44: Is this documented in an issue somewhere?

.github/workflows/build-test
```diff
@@ -52,13 +100,8 @@ then
     python -m pip install --upgrade mypy
 fi
 for MODULE in $MODULES_TO_TEST; do
-    # pytket-iqm not supported with 3.7 or 3.8
-    if [[ "${MODULE}" != "pytket-iqm" || "${PYVER}" == "3.9" ]]; then
-        # FIXME see
-        # https://github.com/CQCL/pytket-extensions/issues/180 and
-        # https://github.com/microsoft/iqsharp/issues/512
-        if [[ "${MODULE}" != "pytket-qsharp" || "${PLAT}" != "Windows" || "${PYVER}" == "3.7" ]]
```
cqc-melf, line 60: Should we close the related issue now?

.github/workflows/build_and_test.yml
```diff
@@ -55,34 +55,29 @@ jobs:
...
+    - name: Set up Python 3.10
       uses: actions/setup-python@v2
       with:
-        python-version: 3.9
-    - name: Build and test (3.9)
-      if: github.event_name == 'push' || github.event_name == 'pull_request'
-      run: |
-        ./.github/workflows/build-test nomypy
-    - name: Build, test and save (3.9)
-      if: github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel')
```
cqc-melf, line 85: what as happened to this release?

modules/pytket-qiskit/docs/intro.txt
```diff
@@ -13,8 +13,8 @@ representations, simulation and access to the `IBMQ <https://www.research.ibm.co
 run on IBM backends and simulators, as well as conversion to and from Qiskit
 representations.
 
-``pytket-qiskit`` is available for Python 3.7, 3.8 and 3.9, on Linux, MacOS and Windows. To
-install, run:
+``pytket-qiskit`` is available for Python 3.8, 3.9 and 3.10, on Linux, MacOS and
```
cqc-melf, line 16: Is there a known problem with M1 macs? Should we bring this up here? Or is that solved?

cqc-melf (commented):
I have a few questions, please take a look on my comments.

.github/workflows/build-test
cqc-alec: No, but I will raise one so that we keep track of it.

.github/workflows/build-test
cqc-alec: No, but I will raise one.

.github/workflows/build-test
cqc-alec: As far as I know the issue is still open. We can build pytket-qsharp on Windows with 3.8 and 3.9, but we cannot test it (at least on the CI, though it seems to work in some local installations). With 3.10 we can't even build it. So unfortunately we can't test pytket-qsharp on Windows on the CI. However I didn't want to remove support altogether because there may be people using it, we can still test locally, and the issue may get fixed.

.github/workflows/build_and_test.yml
cqc-alec: Nothing, this is just a logical simplification of the workflow. The two steps, each conditionally executed, are combined into one, unconditionally executed.

modules/pytket-qiskit/docs/intro.txt
cqc-alec: There is, but it is a qiskit issue. We don't mention M1 Mac support here, though it would probably be worth having a page somewhere explaining the known issues.

.github/workflows/build_and_test.yml
cqc-melf: Perfect, Thank you for the explanation!

.github/workflows/build-test
cqc-melf: Yes, you are right. I forgot that we had a problem with 3.8 and 3.9, too. We should keep it.

modules/pytket-qiskit/docs/intro.txt
cqc-melf: I think it would be helpful to document that somewhere. Not sure what the best place is to do that. Maybe somewhere central in the pytket docs for all known problems?

cqc-melf (approved):

PR #270, opened 2022-01-18
Update Backend objects to use rebase_pass abstract method

sjdilkes:
A consistent part of compiling to different hardware is rebasing a circuit i.e. rewriting each gate in a circuit so that is a (sequence) of gates with OpType permitted by the hardware.

Currently this is completed inconsistently across all Backend. Each Backend now has a `rebase_pass` method which returns a `BasePass` object for applying to a Circuit to complete this transformation. This is also a new abstractmethod of the base `Backend` class.


cqc-alec:
Let's wait before merging this. After https://github.com/CQCL/tket/pull/176 is fixed and merged we can make a dev release of pytket and then update this PR to depend on that -- in case it causes any issues.

cqc-alec (approved):

PR #289, opened 2022-02-08
Update IBMQEmulatorBackend.rebase_pass

sjdilkes:
`IBMQEmulatorBackend` inherits from `AerBackend`, meaning it inherits `AerBackend.rebase_pass` which supports a set of gates that does not match `IBMQEmulatorBackend.required_predicates`.
Note that `IBMQEmulatorBackend.default_compilation_pass` & `IBMQEmulatorBackend.compile_circuit` still support valid circuits as they manually reassign these methods from a generated `IBMQBackend` object.

### DIFF
--- a/modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibmq_emulator.py
+++ b/modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibmq_emulator.py
@@ -34,7 +34,6 @@ from pytket.extensions.qiskit.qiskit_convert import tk_to_qiskit
 from pytket.extensions.qiskit.result_convert import (
     qiskit_experimentresult_to_backendresult,
 )
-from pytket.passes import BasePass  # type: ignore
 from pytket.utils import prepare_circuit
 from pytket.utils.results import KwargTypes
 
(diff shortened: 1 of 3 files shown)


modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibmq_emulator.py
```diff
@@ -46,7 +47,6 @@
 
 if TYPE_CHECKING:
     from pytket.predicates import Predicate  # type: ignore
-    from pytket.passes import BasePass  # type: ignore
```
ss2165, line 49: this line  can remain here as far as i can tell, just need quotes around the annotation

modules/pytket-qiskit/pytket/extensions/qiskit/backends/ibmq_emulator.py
```diff
@@ -100,6 +100,9 @@ def characterisation(self) -> Optional[Dict[str, Any]]:
     def backend_info(self) -> BackendInfo:
         return self._ibmq.backend_info
 
+    def rebase_pass(self) -> BasePass:
```
ss2165, line 103:
```suggestion
    def rebase_pass(self) -> "BasePass":
```

PR #292, opened 2022-02-09
Replace HQS rebase and squash with custom passes

ss2165:
replicates tket definitions of rebase and squash passes e.g.:

https://github.com/CQCL/tket/blob/cc5588ef437107c48c4c4356bccbea4d9f941b01/tket/src/Transformations/Rebase.cpp#L257-L261

### DIFF
--- a/modules/pytket-honeywell/pytket/extensions/honeywell/backends/rebase.py
+++ b/modules/pytket-honeywell/pytket/extensions/honeywell/backends/rebase.py
@@ -1,7 +1,7 @@
 from typing import Set, Union
-from sympy import Expr
+from sympy import Expr  # type: ignore
 from pytket.circuit import OpType, Circuit  # type: ignore
-from pytket.passes import BasePass, RebaseCustom  # type: ignore
+from pytket.passes import SquashCustom, RebaseCustom  # type: ignore
 from pytket.transform import Transform  # type: ignore
 
 
@@ -31,7 +31,7 @@ def _tk1_to_PhasedXRz(
     return circ
 
 
-def _quantinuum_rebase(gate_set: Set[OpType]) -> BasePass:
+def _quantinuum_rebase(gate_set: Set[OpType]) -> RebaseCustom:
     # CX replacement
     c_cx = Circuit(2)
     c_cx.Rz(1.5, 0).Rx(0.5, 1).Rz(1.5, 1).Rx(1.5, 1)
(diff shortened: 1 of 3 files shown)


modules/pytket-honeywell/pytket/extensions/honeywell/backends/rebase.py
```diff
@@ -0,0 +1,55 @@
...
+        circ.Rz(a + b + c, 0)
+    else:
+        circ.Rz(a + c, 0)
+        circ.add_gate(OpType.PhasedX, [b, a], [0])
+
+    Transform.RemoveRedundancies().apply(circ)
+    return circ
+
+
+def _quantinuum_rebase(gate_set: Set[OpType]) -> RebaseCustom:
```
cqc-alec, line 34: Should we check and raise an exception if `gate_set` does not include {Rz, PhasedX, ZZMax}, since the rebase pass won't be correct otherwise? (Then we'd just set `multiops` and `singleops` to these gates below.)

modules/pytket-honeywell/pytket/extensions/honeywell/backends/rebase.py
```diff
@@ -0,0 +1,55 @@
...
+    # CX replacement
+    c_cx = Circuit(2)
+    c_cx.Rz(1.5, 0).Rx(0.5, 1).Rz(1.5, 1).Rx(1.5, 1)
```
cqc-alec, line 37: Shouldn't this use PhasedX and not Rx?

cqc-alec (changes requested):

modules/pytket-honeywell/pytket/extensions/honeywell/backends/rebase.py
```diff
@@ -0,0 +1,55 @@
+from typing import Set, Union
+from sympy import Expr  # type: ignore
+from pytket.circuit import OpType, Circuit  # type: ignore
+from pytket.passes import SquashCustom, RebaseCustom  # type: ignore
+from pytket.transform import Transform  # type: ignore
+
+
+def _approx_0_mod_2(x: Union[float, Expr], eps: float = 1e-10) -> bool:
```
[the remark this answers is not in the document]
ss2165: It's not particularly verbose and I think it is sufficiently confusing as to be worth keeping explicit wherever both types have to be handled correctly.

modules/pytket-honeywell/pytket/extensions/honeywell/backends/rebase.py
ss2165: Though as implemented that's not actually what's happening, I need to fix that

ss2165: https://github.com/CQCL/pytket-extensions/blob/1271f18fb92404b2677df74c081ad1486252ce44/modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py#L150-L156

this is what I actually want

ss2165:
Closing until `RebaseCustom` interface is changed in the base backage

PR #309, opened 2022-02-28
Remove default batch params if submitting to HQS family device.

vanyae-cqc:
HQS (and pytket-honeywell) now allow users to submit to a 'device family' ("HQS-LT") instead of specifying the exact device they want to run on.

We checked with the developers of the HQS API and it turns out that if jobs are submitted to a device family and batching is requested in the request JSON then an error will be returned (as the circuits may be run on different machines).

This PR removes the insertion of default batch params in the case that the user is submitting to a device family. If a user manually inserts batch params then they the API will still return an error, which I figured should be the correct behaviour.

### DIFF
--- a/modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
+++ b/modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
@@ -59,6 +59,7 @@ from .credential_storage import PersistentStorage
 
 _DEBUG_HANDLE_PREFIX = "_MACHINE_DEBUG_"
 HONEYWELL_URL_PREFIX = "https://qapi.honeywell.com/"
+DEVICE_FAMILY = "HQS-LT"
 
 _STATUS_MAP = {
     "queued": StatusEnum.QUEUED,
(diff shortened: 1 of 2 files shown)


modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
```diff
@@ -363,10 +364,13 @@ def process_circuits(
 
             if final_index > 0:
                 # only set batch fields if more than one job submitted
-                body["batch-exec"] = batch_exec
-                if i == final_index:
-                    # flag to signal end of batch
-                    body["batch-end"] = True
+                if (self._device_name != DEVICE_FAMILY) or ("max_batch_cost" in kwargs):
```
ss2165, line 367: The only suggestion I would make is `and`ing this condition with the `if final_index >0` so there is a single decision point as to whether to batch or not.

ss2165 (dismissed):
Just a minor suggestion, looks good otherwise.

modules/pytket-honeywell/pytket/extensions/honeywell/backends/honeywell.py
vanyae-cqc: Yeah that makes sense, will do!

vanyae-cqc:
Merged my fork with Alec's recent merge. He wasn't sure why the MacOs action timed out but suggested running it again. Fingers crossed!

PR #310, opened 2022-03-01
Require pytket 1.0.0rc14

cqc-alec:
Fix up a pytket-qiskit test.

Also add "region" argument to `BraketBackend` constructor and fix up tests following retirement of Aspen-11.

(I tried adding a test for the new OQC device, but it didn't like the ARN; this will need to be investigated separately.)

### DIFF
--- a/modules/pytket-braket/setup.py
+++ b/modules/pytket-braket/setup.py
@@ -37,7 +37,11 @@ setup(
     license="Apache 2",
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
-    install_requires=["pytket == 1.0.0rc14", "amazon-braket-sdk ~= 1.0"],
+    install_requires=[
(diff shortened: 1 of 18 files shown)


modules/pytket-braket/setup.py
```diff
@@ -37,7 +37,11 @@
     license="Apache 2",
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
-    install_requires=["pytket == 1.0.0rc12", "amazon-braket-sdk ~= 1.0"],
+    install_requires=[
+        "pytket == 1.0.0rc14",
+        "amazon-braket-sdk~=1.16",
+        "amazon-braket-schemas~=1.7",
```
cqc-melf, line 43: Should this requirement change be mention in the changelog?

cqc-melf (approved):

PR #312, opened 2022-03-01
Move new test into separate file.

cqc-alec:
Otherwise pytest hangs on the CI with MacOS and Python 3.10.


ss2165 (changes requested):
worth renaming the file ("family_test.py"?) or adding a big comment to say why the split exists so we don't cause a regression

ss2165 (approved):

Issue #258, opened 2022-01-05
[pytket-qiskit] Conversion of qiskit circuit to pytket circuit with non-real angles


cqc-alec:
It is possible to construct a circuit in qiskit which can be successfully converted with `qiskit_to_tk` to a pytket circuit, but where the pytket circuit contains invalid data. The following script (`bad.py`) illustrates this:

```python
from qiskit.opflow import PauliTrotterEvolution
from qiskit.opflow.primitive_ops import PauliSumOp
from pytket.extensions.qiskit import qiskit_to_tk
from pytket.passes import DecomposeBoxes

op = PauliSumOp.from_list([("Z", 1j)])
evolved_op = op.exp_i()
evolution_circ = PauliTrotterEvolution(reps=1).convert(evolved_op).to_circuit()

tk_circ = qiskit_to_tk(evolution_circ)

DecomposeBoxes().apply(tk_circ)

cmds = tk_circ.get_commands()
assert len(cmds) == 1
op = cmds[0].op

print(op.type)
print(op.params)
```

Output:

```
OpType.Rz
Traceback (most recent call last):
  File "bad.py", line 19, in <module>
    print(op.params)
RuntimeError: Not Implemented
```

Roland-djee:
Any thoughts on this @cqc-alec ? What is weird is that in the former case, when the return of `get_commands()` is assigned to a variable, there is no runtime error.

cqc-alec:
The `Not Implemented` just means that there was an invalid (non-real) parameter in the operation. TKET is correct to raise an error here (though the error message could be more helpful). The problem is that an exception should have been thrown earlier: I think it shouldn't be possible to construct such a circuit. i.e. `qiskit_to_tk` should have failed.

cqc-alec:
Because it's not physically meaningful: U3(a,b,c) is non-unitary unless we restrict a, b and c to real values. So we should not accept them as valid parameters.

Roland-djee:
Sure, but none of the parameters from the qiskit `U3` are complex when passed to `qiskit_to_tk`.

Roland-djee:
Here is the output of the same script above where I print `evolution_circ` in two cases where the parameter is `1j` and `1` respectively. In both cases, the circuit parameters are identical but somehow there is an error thrown from pytket when it is complex. It returns the right value otherwise.

```python
❯ python bad.py
evolution circ
    ┌──────────────┐
q: ┤ exp(-i Z)(1) ├
[...]

PR #313, opened 2022-03-02
[TKET-1749] Fix circuit parameter conversion bug when complex.

Roland-djee:
Fixes [TKET-1749](https://cqc.atlassian.net/browse/TKET-1749?atlOrigin=eyJpIjoiM2MzN2Q2YTRlM2EzNDJlZGI3ODE3MWJjNjYxNzJjMDAiLCJwIjoiaiJ9)

This PR fixes the bug reported in #258 and adds exception being thrown when a coefficient in a Pauli string is non-real.


cqc-alec (approved):
LGTM. I think there is also an issue in TKET (that it should not accept the non-real parameters when constructing the circuit), but I will raise a separate tket for that.

PR #324, opened 2022-03-08
Build with 3.9, not 3.10, in release workflow.

cqc-alec:
Because 3.9 is the only version where all extensions are supported.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -65,19 +65,21 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -65,19 +65,21 @@ jobs:
       run: |
         ./.github/workflows/build-test nomypy
     - name: Set up Python 3.9
-      if: github.event_name == 'pull_request'
+      if: github.event_name == 'pull_request' || github.event_name == 'release' || contains(github.ref, 'refs/heads/wheel')
```
cqc-melf, line 68: do we still need the wheel/* branch check here?

cqc-alec: Yeah, so that we can build wheels from a `wheel/**` branch.

cqc-melf (approved):

PR #326, opened 2022-03-09
Update to pytket 1.0.0rc17.

cqc-alec:


cqc-alec:
Ah, sympy update to 1.10 means type annotations need fixing.

ss2165 (approved):

PR #330, opened 2022-03-09
Add support for density-matrix simulators

### LINKED ISSUE #311: [pytket-braket] Error constructing `BraketBackend` with `dm1` simulator
Due to unsupported operation types.

cqc-alec:
Fixes #311 .

Had to resort to a bit of hackery to extract the density matrix from the result, as the obvious method doesn't seem to work, and the internal format is actually different for different devices. Sorry.


yao-cqc (commented):
I'm not sure if the hack is needed. For the test case, I'm able to get it working by using
```
            kwargs["density_matrix"] = result.get_value_by_result_type(
                ResultType.DensityMatrix(target=[0, 1])
            )
```
If the target is not specified, the result type has a internal representation `{'targets': None, 'type': <Type.densitymatrix: 'densitymatrix'>}` but the `GateModelQuantumTaskResult` actually holds an entry `{'targets': [0, 1], 'type': <Type.densitymatrix: 'densitymatrix'>}` hence the KeyError.

cqc-alec:
> I'm not sure if the hack is needed. For the test case, I'm able to get it working by using
> 
> ```
>             kwargs["density_matrix"] = result.get_value_by_result_type(
>                 ResultType.DensityMatrix(target=[0, 1])
>             )
> ```
> 
> If the target is not specified, the result type has a internal representation `{'targets': None, 'type': <Type.densitymatrix: 'densitymatrix'>}` but the `GateModelQuantumTaskResult` actually holds an entry `{'targets': [0, 1], 'type': <Type.densitymatrix: 'densitymatrix'>}` hence the KeyError.

You're right, we can do it this way. Annoyingly, both the task metadata and the format of the density matrix are different depending on whether this is a remote or a local simulator, but I think we can work around this.

yao-cqc (approved):
Looks good! Thanks

PR #333, opened 2022-03-10
fix braket task failure status reason

dhaycraft:
Result object is None if the task fails.

If fail would like it to return the failureReason from Braket

https://github.com/aws/amazon-braket-sdk-python/blob/v1.16.0/src/braket/aws/aws_quantum_task.py#L286-L288

Also is the same in latest versions as well


cqc-alec (approved):
Thanks!

PR #357, opened 2022-04-01
[pytket-quantinuum] allow batches to be resumed in new process_circuits

ss2165:
add two new keyword arguments to process_circuits:
* one to stop a batch being closed
at the end of the list of circuits.
* another to continue using  a previous batch_id

also prevent batching on Syntax Checkers where this is not supported.


cqc-alec (commented):
Would it make sense to have a `get_batch_id(handle)` method on the backend, rather than rely on it being the zeroth element of the handle, which feels a little unstable?

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -299,7 +309,12 @@ def process_circuits(
           perform noisy simulation with an error model (default value is `True`).
         * `group`: string identifier of a collection of jobs, can be used for usage
           tracking.
-        * `max_batch_cost`: maximum HQC usable by submitted batch, default is 500.
+        * `max_batch_cost`: maximum HQC usable by submitted batch, default is
+          500.
+        * `batch_id`: first jobid (first element of ResultHandle) of the batch
```
cqc-alec, line 314: Better to refer to the method here than "first element of ResultHandle".

modules/pytket-quantinuum/tests/api1_test.py
```diff
@@ -81,3 +81,56 @@ def test_device_family(
...
+        circuits=[circ, circ], n_shots=10, valid_check=False, close_batch=False
+    )
+
+    submitted_json = {}
+    if requests_mock.last_request:
+        print(requests_mock.last_request)
+        submitted_json = requests_mock.last_request.json()
+
+    assert "batch-exec" in submitted_json
+    assert submitted_json["batch-exec"] == h1[0]
```
cqc-alec, line 126:
```suggestion
    assert submitted_json["batch-exec"] == backend.get_jobid(h1)
```

modules/pytket-quantinuum/tests/api1_test.py
```diff
@@ -81,3 +81,56 @@ def test_device_family(
...
+    assert "batch-end" not in submitted_json
+
+    _ = backend.process_circuit(
+        circ, n_shots=10, valid_check=False, batch_id=backend.get_jobid(h1)
+    )
+
+    if requests_mock.last_request:
+        submitted_json = requests_mock.last_request.json()
+    assert submitted_json["batch-exec"] == h1[0]
```
cqc-alec, line 135:
```suggestion
    assert submitted_json["batch-exec"] == backend.get_jobid(h1)
```

cqc-alec (changes requested):
I got an error from `test_batching()`:
```
pytket.extensions.quantinuum.backends.api_wrappers.QuantinuumAPIError: HTTP error submitting job, {'code': 18, 'message': 'Batch does not exist'}
```

cqc-alec (approved):

PR #359, opened 2022-04-01
fixed bug in quantinuum user name config

mlk621:
Tom found this error. Small fix.


mlk621:
Seyon has reviewed, there are other things going on. Some other updates may be required on the backend, per Tom, but not this.

PR #362, opened 2022-04-05
Enable pytket extensions remote tests in CI

ferbetanzo:
This PR enables runnig remote tests in the `pull request` flow of the buid_and_test workflow, for the following pytket extensions:
- `pytket-braket`
- `pytket-qiskit`
- `pytket-quantinuum`
- `pytket-iqm`

It also includes the following changes:
- Support for running `pytket-qsharp` remote tests (which for now are disabled due to a bug)
- The use of a strategy matrix so each pytket-extension is tested in parallel
- Changes to the braket backend so the `available_devices` receives an optional `aws_session` (needed by the remote tests)


.github/workflows/build-test
```diff
@@ -104,7 +104,12 @@ for MODULE in $MODULES_TO_TEST; do
                 dotnet iqsharp install --user
             fi
 
-            pytest --doctest-modules
+            # Disable the remote tests for pytket-aqt, -ionq, and -qsharp for now
+            if [[ "${MODULE}" = "pytket-aqt" || "${MODULE}" = "pytket-ionq" || "${MODULE}" = "pytket-qsharp" ]]; then
```
cqc-alec, line 108: The pytket-qsharp remote tests are now working for me. Can we un-disable them here?

modules/pytket-qsharp/tests/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    # If any of the env variables does not exist, then we will try to create
+    # a backend using the local config settings
+    #
+    # Note: by default, the target is the 'ionq_sumulator'
+    env_vars = (
+        "PYTKET_REMOTE_QSHARP_RESOURCE_ID",
+        "PYTKET_REMOTE_QSHARP_LOCATION",
+        "PYTKET_REMOTE_QSHARP_STORAGE",
+    )
+    required_env_vars = [os.getenv(var, default=None) for var in env_vars]
```
cqc-alec, line 42: `default=None` is redundant.

cqc-alec: I don't think we need to create a list here, it's only used in the `any` below.

modules/pytket-qsharp/tests/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    if any(var is None for var in required_env_vars):
```
cqc-alec, line 43:
```suggestion
    if any(os.getenv(var) is None for var in env_vars):
```

.github/workflows/build_and_test.yml
```diff
@@ -16,43 +16,82 @@ on:
...
         fetch-depth: '0'
     - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/* +refs/heads/*:refs/remotes/origin/*
-    - name: Initialize modules list
-      run: |
-        cd ${GITHUB_WORKSPACE}/modules
-        LS_MODULES=`ls -d */`
-        ALL_MODULES=`echo ${LS_MODULES} | sed -e 's/\///g'`
-        echo "MODULES_TO_TEST=${ALL_MODULES}" >> $GITHUB_ENV
-    - name: check for mofified modules
+    - name: Only test modified modules for PRs
```
cqc-alec, line 78: I don't think this is working. For instance, in this PR we see pytket-aqt being tested, even though it is unmodified. We seem to be setting `MODULES_TO_TEST` in the environment to `matrix.module`, overwriting this value.

cqc-alec (changes requested):

.github/workflows/build-test
```diff
@@ -104,7 +104,12 @@ for MODULE in $MODULES_TO_TEST; do
...
+                unset PYTKET_RUN_REMOTE_TESTS; pytest --doctest-modules
+            else
+                pytest --doctest-modules
+            fi
```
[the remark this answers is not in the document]
ferbetanzo: The problem with doing `PYTKET_RUN_REMOTE_TESTS=0` is that in the python code, if you do `os.getenv("VAR")`, it will only return None if the variable does not exist. By setting it to 0, then the code for the remote tests will execute as the value is not None but 0.

Now, it's true that doing the `unset PYTKET_RUN_REMOTE_TESTS` will leave the var unset, but that's only for the rest of this bash script. It will not modify the value of the variable from the caller. Since we're now executing the script just once per module (as we're running them in parallel), I don't think this is too much problem.

Still, I could save the value of the variable and restore it later after running the pytest. Should I apply that change?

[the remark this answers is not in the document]
cqc-alec: I think we should either adjust the python code so that it treats the value 0 equivalently to None (unset), or modify this script so that it clearly only tests one module: read from a variable like `MODULE`, and get rid of the loops `for MODULE in $MODULES_TO_TEST`. Otherwise it's very confusing since the script is obviously written to work with a list of modules but doesn't.

cqc-alec:
Note that the three checks "Build and test (Linux)" etc are "required" for this branch, but with these changes they no longer exist. Should they be "dummy" checks depending on successful completion of all others, or is there a better way to make all relevant sub-checks "required"?

ferbetanzo:
> Note that the three checks "Build and test (Linux)" etc are "required" for this branch, but with these changes they no longer exist. Should they be "dummy" checks depending on successful completion of all others, or is there a better way to make all relevant sub-checks "required"?

This seems to be a bug in GH (see [this](https://stackoverflow.com/questions/52200096/github-pull-request-waiting-for-status-to-be-reported)). I'll push the changes with your suggestions and see if that helps :)

ferbetanzo:
@cqc-alec , I think I've handled all your feedback except enabling the pytket-qsharp remote tests. Since that will take me longer, and also there's already a ticket for doing just that, would you be ok if we keep it disabled for now? That would of course only be the remote tests, the other ones that do not require a remote backend would still be run.

Also... I'm getting a crash from pylint in the CI, but locally it works fine for me. Have you ever faced anything similar?

cqc-alec:
> @cqc-alec , I think I've handled all your feedback except enabling the pytket-qsharp remote tests. Since that will take me longer, and also there's already a ticket for doing just that, would you be ok if we keep it disabled for now? That would of course only be the remote tests, the other ones that do not require a remote backend would still be run.

OK, sure.

> Also... I'm getting a crash from pylint in the CI, but locally it works fine for me. Have you ever faced anything similar?

Ah, this is the same issue we had on tket, fixed by https://github.com/CQCL/tket/pull/342 -- I'll do a similar PR here.

cqc-alec:
> Ah, this is the same issue we had on tket, fixed by [CQCL/tket#342](https://github.com/CQCL/tket/pull/342) -- I'll do a similar PR here.

Now merged -- I think if you close and reopen the PR (or just do a merge from `develop`) it will pick up the fix.

ferbetanzo:
> > Ah, this is the same issue we had on tket, fixed by [CQCL/tket#342](https://github.com/CQCL/tket/pull/342) -- I'll do a similar PR here.
> 
> Now merged -- I think if you close and reopen the PR (or just do a merge from `develop`) it will pick up the fix.

I'll close and re-open the PR to see if that fixes also the "Expected - Waiting for status to be reported".

ferbetanzo:
In order to fix the pending checks, I think we need to remove the "outdated" checks in the branch settings (according to [this](https://github.community/t/github-actions-pull-requests-always-include-some-outdated-checks/16157)), and then we'd need to add each of the pytket-ext check for each platform as required checks. Otherwise, I think I can leave the "Build and test (platform)" jobs as required checks, and make them depend on new matrix check jobs.

Which do you reckon is better?

cqc-alec:
> In order to fix the pending checks, I think we need to remove the "outdated" checks in the branch settings (according to [this](https://github.community/t/github-actions-pull-requests-always-include-some-outdated-checks/16157)), and then we'd need to add each of the pytket-ext check for each platform as required checks. Otherwise, I think I can leave the "Build and test (platform)" jobs as required checks, and make them depend on new matrix check jobs.
> 
> Which do you reckon is better?

I don't mind changing the set of required checks in github settings, but presumably then it would have to include checks corresponding to modules that may not be tested (if they are unmodified). I would also prefer not to have to update either the workflow script or the github settings when a new module is added. So can we restore jobs with the old names that depend on successful completion of the modules-to-test?

ferbetanzo:
> So can we restore jobs with the old names that depend on successful completion of the modules-to-test?

Done :)

cqc-alec (approved):

PR #380, opened 2022-04-21
[pytket-braket] Handle measurements.

### LINKED ISSUE #375: [pytket-braket] get_compiled_circuit step adds additional qubits when mapping device for Aspen-M-1 device
Was testing out changes for #348 and found that when using get_compiled_circuit will map to more qubits than were in the original problem for Aspen-M-1 device for certain circuits. I guess this isn't super surprising since the connectivity graph is kind of sparse so any time there is dense connectivity in your circuit need to add additional bits. However this makes the results being read out from the machine a different dimension than the original logical qubits for the circuit that was compiled. Is there any sort of reasonable way to select qubits to map back to the original logical qubits or to aggregate from physical qubits?

cqc-alec:
Fixes #375 .

Currently we ignore measurement operations in tket circuits, since braket automatically measures everything. But we need to be able to handle circuits where not all qubits are measured. Here we keep track of the measurement mappings in a dictionary encoded in the result handle.

### DIFF
--- a/modules/pytket-braket/pytket/extensions/braket/backends/braket.py
+++ b/modules/pytket-braket/pytket/extensions/braket/backends/braket.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+from ast import literal_eval
 from enum import Enum
 import time
 from typing import (
@@ -204,6 +205,7 @@ def _obs_from_qpo(operator: QubitPauliOperator, n_qubits: int) -> Observable:
 def _get_result(
     completed_task: Union[AwsQuantumTask, LocalQuantumTask],
     target_qubits: str,
+    measures: Dict[int, int],
     want_state: bool,
     want_dm: bool,
     ppcirc: Optional[Circuit] = None,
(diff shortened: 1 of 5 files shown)


modules/pytket-braket/pytket/extensions/braket/backends/braket.py
```diff
@@ -203,21 +203,21 @@ def _obs_from_qpo(operator: QubitPauliOperator, n_qubits: int) -> Observable:
 
 def _get_result(
     completed_task: Union[AwsQuantumTask, LocalQuantumTask],
-    target_qubits: str,
+    target_qubits: List[int],
+    measures: Dict[int, int],
```
sjdilkes, line 207: Can we assume from Braket that qubits will always be single indexed and won't have different strings to describe different subsets of qubits?

sjdilkes (changes requested):
Just a single question.

modules/pytket-braket/pytket/extensions/braket/backends/braket.py
cqc-alec: Good question. No, but it is already the case that `tk_to_braket` assumes all qubits belong to a single, singly-indexed, register. So we are not breaking anything here, but I will see if I can relax that assumption, and if not add a predicate to assert that we only support such circuits.

sjdilkes (approved):

Issue #388, opened 2022-05-11
[pytket-braket] Remote tests failing on CI


cqc-alec:
See https://github.com/CQCL/pytket-extensions/runs/6384025167
I can run them successfully locally.

PR #390, opened 2022-05-12
[pytket-braket] add additional cancelling status for braket

dhaycraft:
When doing constant polling for status there is an additional status which is not handled by StatusEnum.Cancelled 
Which ends up being handled by this [line](https://github.com/CQCL/pytket-extensions/blob/develop/modules/pytket-braket/pytket/extensions/braket/backends/braket.py#L768)
`Unrecognized state 'CANCELLING'`

Let me know if you'd like other behavior in handling this status.
Thanks!


cqc-alec:
Thanks for this. Tests are failing for unrelated reasons (https://github.com/CQCL/pytket-extensions/issues/388); will get this merged as soon as that is fixed.

cqc-alec (approved):

PR #406, opened 2022-05-26
Ensure MODULE environment variable set in hotfix workflow

cqc-alec:


ferbetanzo (approved):
Sorry, I totally missed the hotfix.yml workflow. Thx for fixing it :)

PR #419, opened 2022-06-13
Add group argument to QuantinuumBackend

yao-cqc:


modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -371,6 +376,8 @@ def process_circuits(
         group = kwargs.get("group")
```
ss2165, line 376:
```suggestion
        group = kwargs.get("group", self._group)
```

then the new elif block can be removed

yao-cqc: nice!

PR #420, opened 2022-06-14
[pytket-quantinuum] Catch ValueError for uvloop

johnchildren:
When using nest_asyncio in the context of alternative event loops such as `uvloop`, `nest_asyncio.apply()` will raise a ValueError rather than a RuntimeError. For example:

```
ValueError: Can't patch loop of type <class 'uvloop.Loop'>
```

This commit adds ValueError to the `except` block in pytket-quantinuum to ignore this error and continue.


johnchildren:
https://github.com/MagicStack/uvloop/issues/405 There is an issue for supporting `nest_asyncio` in `uvloop`, but it seems unlikely that someone running pytket-quantinuum with uvloop as the current event loop would want to apply the patch.

cqc-alec (approved):

PR #421, opened 2022-06-15
update doc build

cqc-melf:
fix error in the docbuild

### DIFF
--- a/modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -372,7 +372,7 @@ class QuantinuumBackend(Backend):
         * `close_batch`: boolean flag to close the batch after the last circuit,
            default=True.
         * `wasm_file_handler`: a ``WasmFileHandler`` object for linked WASM module.
-        
+
         """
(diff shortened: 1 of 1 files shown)


modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -372,6 +372,7 @@ def process_circuits(
         * `close_batch`: boolean flag to close the batch after the last circuit,
            default=True.
         * `wasm_file_handler`: a ``WasmFileHandler`` object for linked WASM module.
+
```
cqc-alec, line 375: Is this line needed?

cqc-melf: Yes, if it is missing the doc build complains... not sure why...

cqc-alec (approved):

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
cqc-melf: Might be possible to replace the empty line by an `\n` at the end of the line above, if you would prefer that I could try.

PR #422, opened 2022-06-16
[pytket-qiskit] replace bogota with manila

cqc-melf:


cqc-melf:
closed because of a pytket bug, will reopen after the bug is fixed

cqc-melf:
closed because of a pytket bug, will reopen after the bug is fixed

PR #424, opened 2022-06-16
[pytket-qiskit] Fix test.

cqc-alec:


modules/pytket-qiskit/tests/backend_test.py
```diff
@@ -993,7 +993,11 @@ def test_compilation_correctness(bogota_backend: IBMQBackend) -> None:
         m_ini = lift_perm(ini)
         m_inv_fin = lift_perm(inv_fin)
 
-        assert compare_unitaries(u, m_inv_fin @ compiled_u @ m_ini)
+        # Note that we do not expect unitary equivalence, because the pass includes
```
cqc-melf, line 996: Is this something that has changes between the pytket version?

cqc-alec: The test has been wrong ever since we introduced contextual optimization. But it just happened to pass, until [this](https://github.com/CQCL/tket/pull/401) improvement to the rebase pass caused it to fail.

cqc-melf: Good that we have find that :)

cqc-melf (approved):

PR #439, opened 2022-07-06
add pytket pass submission

cqc-melf:


modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -407,6 +409,14 @@ def process_circuits(
                 "utf-8"
             )
 
+        try:
+            pytket_pass = cast(BasePass, kwargs.get("pytketpass"))
+        except ValueError:
+            pytket_pass = None
+
+        if pytket_pass is not None:
+            basebody["options"]["tket-pass"] = pytket_pass.to_dict()
```
cqc-alec, line 418: For future extensions I think we should have a `tket` field under `options`, with further subfields, of which `compilation-pass` can be one. So:
```
options:
    tket:
        compilation-pass: ...
        ...
    ...
```
instead of
```
options:
    tket-pass: ...
    ...
```

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -407,6 +409,14 @@ def process_circuits(
                 "utf-8"
             )
 
+        try:
+            pytket_pass = cast(BasePass, kwargs.get("pytketpass"))
+        except ValueError:
+            pytket_pass = None
```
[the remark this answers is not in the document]
cqc-melf: I have changes it similar to your suggestion, thank you!

cqc-melf:
> Have you checked that submitting this to an actual device will not raise an error? Maybe in the documentation it should make clear that it may be ignored.

I have check a lot of possible combinations, I was not able to generate an error with this. (even after the change to the "tket" field. I have change my "beta" commend, which was a bit hidden to a more obvious comment. (the backend does not support this for now, that is why I have not added something to the changelog)

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -372,6 +372,8 @@ def process_circuits(
         * `close_batch`: boolean flag to close the batch after the last circuit,
            default=True.
         * `wasm_file_handler`: a ``WasmFileHandler`` object for linked WASM module.
+        * `pytketpass`: a ``pytket.passes.BasePass`` which should be applied
+           by the backend (might be ignored by the backend).
```
cqc-alec, line 376:
```suggestion
        * `pytketpass`: a ``pytket.passes.BasePass`` intended to be applied
           by the backend (beta feature, may be ignored).
```

cqc-alec (changes requested):
Just a rewording suggestion.

cqc-alec (approved):

PR #442, opened 2022-07-13
[pytket-iqm] Restore iqm extension

cqc-alec:


modules/pytket-iqm/tests/backend_test.py
```diff
@@ -18,7 +18,7 @@
 from pytket.circuit import Circuit  # type: ignore
 from pytket.backends import StatusEnum
 from pytket.extensions.iqm import IQMBackend
-from requests import HTTPError, get
+from requests import get
 
 curr_file_path = Path(__file__).resolve().parent
 iqm_demo_url = "https://cortex-demo.qc.iqm.fi/"
```
ferbetanzo, line 24: Would it be better to define these, `curr_file_path` (or the full path to the `demo_settings.json`) and `iqm_demo_url`, just once in the `conftest.py` and then use those values for the authenticated backend fixture and also in the tests?

ferbetanzo (approved):

PR #446, opened 2022-07-14
upgrade qiskit to 0.37

cqc-melf:


modules/pytket-qiskit/pytket/extensions/qiskit/tket_pass.py
```diff
@@ -75,7 +75,7 @@ class TketAutoPass(TketPass):
     }
 
     def __init__(
-        self, backend: Union[BaseBackend, BackendV1], optimisation_level: int = 1
+        self, backend: Union[BackendV1], optimisation_level: int = 1
```
ss2165, line 78: doesn't need to be a union any more

ss2165 (changes requested):

modules/pytket-qiskit/pytket/extensions/qiskit/tket_pass.py
cqc-melf: Thank you, have removed that.

PR #450, opened 2022-07-21
[feature] [pytket-quantinuum] add new simpler batching interface

ss2165:
The Quantinuum batching system is not  a good fit for process_circuits as an interface, and the way of doing it with keyword arguments was getting cumbersome. This simplifies things, at the cost of no longer matching a generic interface. It also fixes an existing bug where one could not start a batch by submitting just a single circuit.

### DIFF
--- a/modules/pytket-quantinuum/docs/intro.txt
+++ b/modules/pytket-quantinuum/docs/intro.txt
@@ -119,13 +119,41 @@ The ``process_circuits`` method for the QuantinuumBackend accepts the following 
 *  ``noisy_simulation`` : boolean flag to specify whether the simulator should
    perform noisy simulation with an error model (default value is ``True``).
 *  ``group`` : string identifier of a collection of jobs, can be used for usage tracking.
-*  ``max_batch_cost``: maximum HQC usable by submitted batch, default is 500.
-*  ``batch_id`` : first ``jobid`` of the batch to which this batch of circuits should be submitted. Job IDs can be retrieved from ResultHandle using ``backend.get_jobid(handle)``.
-*  ``close_batch`` : boolean flag to close the batch after the last circuit in the job (default value is ``True``).
 
 For the Quantinuum ``Backend``, ``process_circuits`` returns a ``ResultHandle`` object containing a ``job_id`` and a postprocessing ( ``ppcirc``) circuit if there is one.
 
 The ``logout()`` method clears stored JSON web tokens and the user will have to sign in again to access the Quantinuum API.
+
+Batching
+--------
+Quantinuum backends (except syntax checkers) support batching of jobs (circuits). To create
+a batch of jobs, users submit the first job, then signal that subsequent jobs should
+be added to the same batch using the handle of the first. The backend queue
+management system will start the batch as soon as the first job reaches the
+front of the queue and ensure subsequent batch jobs are run one after the other,
+until the end of the batch is reached or there are no new jobs added to the batch
+for ~1 min (at which point the batch expires and any subsequent jobs will be
+added to the standard queue). 
+
+The standard ``process_circuits`` method **no
+longer batches by default**. To use batching first start the batch with
+``start_batch``, which has a similar interface to ``process_circuit`` but with
+an extra first argument `max_batch_cost`:
+
+::
+  
+  h1 = backend.start_batch(max_batch_cost=300, circuit=circuit, n_shots=100)
+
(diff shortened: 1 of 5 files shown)


modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -483,19 +469,69 @@ def process_circuits(
...
+
+    def start_batch(
+        self,
+        circuit: Circuit,
+        n_shots: Union[None, int] = None,
+        max_batch_cost: int = 500,
+        valid_check: bool = True,
+        **kwargs: QuumKwargTypes,
+    ) -> ResultHandle:
+        """Start a batch of jobs on the backend, behaves like `process_circuit`
```
cqc-melf, line 492: Could you add here some more documentation about the different parameters?

modules/pytket-quantinuum/docs/intro.txt
```diff
@@ -119,14 +119,42 @@ The ``process_circuits`` method for the QuantinuumBackend accepts the following
...
 
+Batching
+--------
+Quantinuum backends (except syntax checkers) support batching of jobs (circuits). To create
+a batch of jobs, users submit the first job, then signal that subsequent jobs should
+be added to the same batch using the handle of the first. The backend queue
+management system will start the batch as soon as the first job reaches the
+front of the queue and ensure subsequent batch jobs are run one after the other,
+until the end of the batch is reached or there are no new jobs added to the batch
+for ~1 min (at which point the batch expires and any subsequent jobs will be
```
cqc-melf, line 135: Does this include an automated submission after 1 minute?

modules/pytket-quantinuum/tests/api1_test.py
```diff
@@ -348,31 +350,26 @@ def test_device_family(
...
-        circuits=[circ, circ], n_shots=10, valid_check=False, **kwargs
-    )
-
-    submitted_json = {}
-    if requests_mock.last_request:
-        submitted_json = requests_mock.last_request.json()
+    circ = backend.get_compiled_circuit(circ)
 
+    max_batch_cost = 20
     if chosen_device == "H1":
```
cqc-melf, line 362: Is it planed to ad a support for this to H1 or will that not happen?

ss2165: this is just to check that check for batching capability is working, if support is added by the API it will automatically be enabled by the backend

modules/pytket-quantinuum/docs/intro.txt
ss2165: I'm not sure what you mean

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -483,19 +469,88 @@ def process_circuits(
...
         return handle_list
 
+    def _check_batchable(self) -> None:
+        if self.backend_info:
+            if not self.backend_info.misc.get("batching", False):
+                raise BatchingUnsupported()
+
+    def start_batch(
+        self,
+        max_batch_cost: int,
```
ss2165, line 486: note max_batch_cost is now a required parameter to reflect some upcoming changes in the API (different organisations having different maxima)

cqc-melf:
It is a bit hard to review if you force push your changes, you can squash your commits at the end when merging to develop.

modules/pytket-quantinuum/docs/intro.txt
cqc-melf: If a batch is started is that directly submitted to the API or is that done later after waiting some time or does that require any action? From my understanding of the code that is directly submitted? In the example you are using the `batch_end=True` is that required?

It might be helpful to outline to the users somewhere on witch cases it is recommended to use `start_batch` and in which the old `process_circuits` ? What do you think?

ss2165: If a batch is started it is immediately submitted to the API. The batch management system is entirely done on the backend. The 1 minute time interval is between batch job submissions to the API. 

I don't really have any solid recommendations about when to use what, which is why I've gone with just stating the system as clearly as possible and letting the user decide.

cqc-melf: Okay, thank you!

PR #454, opened 2022-07-21
[feature][pytket-quantinuum] new direct submit_qasm method

ss2165:
~blocked by #450 , review only the final commit~

### DIFF
--- a/modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
+++ b/modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
@@ -359,6 +359,95 @@ class QuantinuumBackend(Backend):
         """
         return cast(str, handle[0])
 
+    def submit_qasm(
+        self,
+        qasm: str,
+        n_shots: int,
+        name: Optional[str] = None,
+        noisy_simulation: bool = True,
+        group: Optional[str] = None,
+        wasm_file_handler: Optional[WasmFileHandler] = None,
+        pytket_pass: Optional[BasePass] = None,
+        parametrized_zz: bool = False,
+        request_options: Optional[Dict[str, Any]] = None,
+    ) -> ResultHandle:
+        """Submit a qasm program directly to the backend.
+
+        :param qasm: QASM 2.0 program.
+        :type qasm: str
+        :param n_shots: Number of shots
+        :type n_shots: int
+        :param name: Job name, defaults to None
+        :type name: Optional[str], optional
+        :param noisy_simulation: Boolean flag to specify whether the simulator should
+          perform noisy simulation with an error model defaults to True
+        :type noisy_simulation: bool, optional
+        :param group: String identifier of a collection of jobs, can be used for usage
+          tracking. Overrides the instance variable `group`, defaults to None
+        :type group: Optional[str], optional
+        :param wasm_file_handler: ``WasmFileHandler`` object for linked WASM
+            module, defaults to None
+        :type wasm_file_handler: Optional[WasmFileHandler], optional
+        :param pytket_pass: ``pytket.passes.BasePass`` intended to be applied
+           by the backend (beta feature, may be ignored), defaults to None
+        :type pytket_pass: Optional[BasePass], optional
+        :param request_options: Extra options to add to the request body as a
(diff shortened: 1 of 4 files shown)


modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -359,6 +359,95 @@ def get_jobid(self, handle: ResultHandle) -> str:
...
+        parametrized_zz: bool = False,
+        request_options: Optional[Dict[str, Any]] = None,
+    ) -> ResultHandle:
+        """Submit a qasm program directly to the backend.
+
+        :param qasm: QASM 2.0 program.
+        :type qasm: str
+        :param n_shots: Number of shots
+        :type n_shots: int
+        :param name: Job name, defaults to None
```
cqc-alec, line 380: This parameter is unused. (Should go in body of request?)

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -359,6 +359,95 @@ def get_jobid(self, handle: ResultHandle) -> str:
...
+        :param wasm_file_handler: ``WasmFileHandler`` object for linked WASM
+            module, defaults to None
+        :type wasm_file_handler: Optional[WasmFileHandler], optional
+        :param pytket_pass: ``pytket.passes.BasePass`` intended to be applied
+           by the backend (beta feature, may be ignored), defaults to None
+        :type pytket_pass: Optional[BasePass], optional
+        :param request_options: Extra options to add to the request body as a
+          json-style dictionary, defaults to None
+        :type request_options: Optional[Dict[str, Any]], optional
+        :raises WasmUnsupported: Wasm submitted to unsupported backend.
```
cqc-alec, line 397:
```suggestion
        :raises WasmUnsupported: WASM submitted to backend that does not support it.
```

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
```diff
@@ -379,7 +468,6 @@ def process_circuits(
         * `wasm_file_handler`: a ``WasmFileHandler`` object for linked WASM module.
         * `pytketpass`: a ``pytket.passes.BasePass`` intended to be applied
            by the backend (beta feature, may be ignored).
-        * `use_websocket`: boolean flag to allow using a websocket connection.
```
cqc-alec, line 382: Hm, seems this was unused. Yet it is used in the call from the `cost()` function. Is this a bug?

cqc-alec (changes requested):

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
ss2165: It was used when process_circuits was handling batching as it would check on the status of the first job (which is why the retrieve_job calling methods need it). It should have been removed in #450  but was leftover.

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
ss2165: I remember moving that line...must have got deleted somehow, this is the danger with big refactors I guess, nice catch!

ss2165: Have fixed and cleaned up a bit generally

modules/pytket-quantinuum/pytket/extensions/quantinuum/backends/quantinuum.py
cqc-alec: OK. Will its use by the `cost()` method be fixed in a separate PR?

ss2165: no, that was an oversight, fixed now

cqc-alec (approved):

cqc-melf:
@ss2165 do you want this changes to be merged and transferred to the new repo?

PR #455, opened 2022-07-25
Update cirq conversion to support cirq.ops.common_gates rx, ry and rz

sjdilkes:
Previously Cirq only supported such gates as XPowGate, YPowGate and ZPowGate (or we missed the operations). Now they are supported and are also a typical user primitive, update the conversion code to use them. As these gates use radians instead of exponents I've added them as a special case.

Note also {X, Y, Z}PowGate and r{x, y, z} are both mapped to OpType.R{x, y, z} which are always mapped back to {X, Y, Z}PowGate. This means the conversion from cirq -> tk -> cirq is not identical for cases with rotation gates. This seems fine to me given they have two operations that are schematically different but logically the same, but it does mean I've added a new test case for "radian gates" which doesn't expect the cirq circuit strings are the sae.


yao-cqc (approved):
LGTM

PR #459, opened 2022-07-27
Braket nb

CalMacCQ:
This notebook is a bit untidy at the moment, hence I've made it a draft for now. I'm not sure what is/is not helpful to include in a notebook like this. 

If this isn't the right place for a notebook example then I'm happy to remove it.

### DIFF
--- a/modules/pytket-braket/pytket-braket_example.ipynb
+++ b/modules/pytket-braket/pytket-braket_example.ipynb
@@ -662,18 +662,6 @@
    ]
   },
   {
-   "cell_type": "markdown",
-   "id": "f1969647",
-   "metadata": {},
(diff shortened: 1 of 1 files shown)


cqc-alec:
Currently I think the right place for this would be https://github.com/CQCL/pytket/tree/main/examples .
This will require adding pytket-braket to the requirements there, checking in a corresponding python file, and updating `check-examples`.
Alternatively it could live in an `examples` repo here, but should have some kind of test, preferably on the CI,

cqc-melf:
Braket moved over to https://github.com/CQCL/pytket-braket, please reopen the PR at the new repo. If you need help, please let me know.

PR #461, opened 2022-07-27
Feature/improve state unitary retrieval for aer and qulacs

yao-cqc:
``QulacsBackend``, ``AerStateBackend`` and ``AerUnitaryBackend`` now take care of implicit qubit permutation and wire endianness before running circuits. The result retrieval speed is now instant given default permutation and endianness.

### DIFF
--- a/modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py
+++ b/modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py
@@ -66,7 +66,7 @@ from pytket.extensions.qiskit._metadata import __extension_version__
 from pytket.architecture import Architecture  # type: ignore
 from pytket.placement import NoiseAwarePlacement  # type: ignore
 from pytket.utils.operators import QubitPauliOperator
-from pytket.utils.results import KwargTypes, permute_basis_indexing
+from pytket.utils.results import KwargTypes
 from qiskit import Aer  # type: ignore
 from qiskit.providers.aer.library import (  # type: ignore # pylint: disable=unused-import
     save_expectation_value,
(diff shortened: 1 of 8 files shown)


modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py
```diff
@@ -159,13 +159,23 @@ def process_circuits(
...
             elif self._backend_info.device_name == "aer_simulator_unitary":
+                qcs = [
+                    tk_to_qiskit(tkc, reverse_index=True, replace_implicit_swaps=True)
+                    for tkc in batch
+                ]
                 for qc in qcs:
                     qc.save_unitary()
+            else:
+                qcs = [tk_to_qiskit(tkc) for tkc in batch]
+
```
ss2165, line 178:
```suggestion
        if self.supports_state:
            reverse_index, replace_implicit_swaps = (True, True)
            for qc in qcs:
                qc.save_state()

        elif self.supports_unitary:
            reverse_index, replace_implicit_swaps = (True, True)
            for qc in qcs:
                qc.save_unitary()

        qcs = [
            tk_to_qiskit(
                tkc, reverse_index=reverse_index, replace_implicit_swaps=replace_implicit_swaps
            )
            for tkc in batch
        ]


```

ss2165 (changes requested):
Looks good! I think it would be good to have a test of the "replace_implicit_swaps" method in at least one of the backends as a sanity check. You could refactor it out in to a method that returns a list of pairs of integers and test that independently. Then use those integers to apply swaps.

yao-cqc:
As for the code duplication, I think it might be worth to add a pass to convert implicit swaps to actual swaps in tket as well as a method to convert the endianness of the circuit (similar to Qiskit's ``reverse_bits``).

modules/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py
```diff
@@ -158,24 +158,19 @@ def process_circuits(
...
 
+        for (n_shots, batch), indices in zip(circuit_batches, batch_order):
+            qcs = []
+            for tkc in batch:
+                qc = tk_to_qiskit(tkc, reverse_index, replace_implicit_swaps)
+                if self.supports_state:
+                    qc.save_state()
+                elif self.supports_unitary:
+                    qc.save_unitary()
+                qcs.append(qc)
```
ss2165, line 173: nice 👍

ss2165 (approved):

PR #465, opened 2022-07-29
Remove/iqm braket

cqc-melf:
moved to new repo


cqc-alec (approved):

cqc-alec:
Perhaps wait for successful releases from new repo before merging?

cqc-alec:
> Perhaps wait for successful releases from new repos before merging?

On second thoughts, these are the ones that were failing because of remote issues, right? So scratch that.

cqc-melf:
I have made RC releases from the new repos for which I had to disable the remote tests, so it would be possible to to a real release if you want?

cqc-melf:
See https://pypi.org/project/pytket-braket/#history and https://pypi.org/project/pytket-iqm/#history
For pytket iqm it might be interesting to do a real release even with the not working tests, because the version containing pytket .10 has not been released yet. But I still think it would be better to solve the issues before doing that.

cqc-alec:
Yeah, I don't think there's an urgent need to release either yet.

PR #469, opened 2022-08-04
Update build-docs

cqc-melf:


cqc-melf:
@cqc-alec is there any easy way to build and upload the docs without making a release?

cqc-alec:
Not really, this was one motivation for moving to separate repos. But you can edit the existing workflows to do this on a branch.

cqc-melf:
The problem is that we are currently missing the links for braket and iqm docs, I will try to solve this on this branch to add them without a release.

cqc-alec (approved):

PR #471, opened 2022-08-05
fix edge to the same node

cqc-melf:
solves issue in: https://github.com/CQCL/pytket/pull/164

### DIFF
--- a/modules/pytket-qiskit/docs/changelog.rst
+++ b/modules/pytket-qiskit/docs/changelog.rst
@@ -1,6 +1,11 @@
 Changelog
 ~~~~~~~~~
 
+0.28.1 (August 2022)
+--------------------
+
+* Fix error in aer backend creation
+
 0.28.0 (August 2022)
 --------------------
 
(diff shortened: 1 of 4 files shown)


modules/pytket-qiskit/docs/changelog.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 ~~~~~~~~~
 
+0.28.1 (August 2022)
```
cqc-alec, line 4:
```suggestion
0.28.1 (unreleased)
```

cqc-alec (changes requested):
This seems harmless, but I am not sure it is the right fix as it hides the erroneous construction of the `NoiseModel`. It looks like the bug is in the example `entanglement_swapping` notebook, where `make_noise_model` adds errors between identical nodes:
```python
    for i, j in product(qubits, repeat=2):
        model.add_quantum_error(dep_err, ["cx"], [i, j])
```

cqc-melf:
I had not seen that in the notebook, in that case I would close this and fix this in the notebook.

PR #475, opened 2022-08-15
add CRx and CRy

CalMacCQ:
As pointed out in the #tket-support channel `tk_to_qiskit` was failing for CRx and CRy operations. I have now added these to qiskit_convert.py to fix this.

I don't think there is any reason why these were left out but correct me if I'm wrong.

I have moved CRy fron `_qiskit_gates_other` to `_qiskit_gates_2q` is this the right place for it?

There is also a comment here about the CnRy which is probably irrelevant to this.

```
One way mapping: CRY is a special case of CnRy, but not vice versa.
Constructing a qiskit CnRy gate is not just a single step,
so treat as a special case.
del _known_qiskit_gate_rev[OpType.CnRy]
```

### DIFF
--- a/modules/pytket-qiskit/tests/qiskit_convert_test.py
+++ b/modules/pytket-qiskit/tests/qiskit_convert_test.py
@@ -684,7 +684,7 @@ def test_convert_multi_c_reg() -> None:
     assert circ.get_commands()[0].args == [m0, q1]
 
 
-# test that tk_to_qiskit works after adding OpType.CRx and OpType.CRy (TKET-2386)
+# test that tk_to_qiskit works after adding OpType.CRx and OpType.CRy
 def test_crx_and_cry() -> None:
     tk_circ = Circuit(2)
     tk_circ.CRx(0.5, 0, 1)
--- a/modules/pytket-qiskit/tests/qiskit_convert_test.py
+++ b/modules/pytket-qiskit/tests/qiskit_convert_test.py
@@ -682,3 +682,13 @@ def test_convert_multi_c_reg() -> None:
     qcirc = tk_to_qiskit(c)
     circ = qiskit_to_tk(qcirc)
     assert circ.get_commands()[0].args == [m0, q1]
+
+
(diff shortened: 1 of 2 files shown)


CalMacCQ:
Looks like tests are failing. Maybe I need to add a separate OpType for CnRy?

yao-cqc:
> Looks like tests are failing. Maybe I need to add a separate OpType for CnRy? There are some references to CnRy in `append_tk_command_to_qiskit` and `add_qiskit_data`.

I think deleting the following lines will fix the issue.

```
One way mapping: CRY is a special case of CnRy, but not vice versa.
Constructing a qiskit CnRy gate is not just a single step,
so treat as a special case.
del _known_qiskit_gate_rev[OpType.CnRy]
```

Can also you add a test case where `tk_to_qiskit` was failing for CRx and CRy operations?

CalMacCQ:
It seems there are still some builds failing. Perhaps there are issues with the test that I wrote?

CalMacCQ:
I'll call this done for now. Let me know if the test is okay and if there's anything else I should add.

modules/pytket-qiskit/tests/qiskit_convert_test.py
```diff
@@ -682,3 +682,13 @@ def test_convert_multi_c_reg() -> None:
     qcirc = tk_to_qiskit(c)
     circ = qiskit_to_tk(qcirc)
     assert circ.get_commands()[0].args == [m0, q1]
+
+
+# test that tk_to_qiskit works after adding OpType.CRx and OpType.CRy (TKET-2386)
```
yao-cqc, line 687: Can you remove the internal Jira ticket reference?

CalMacCQ: Ah good point, added another commit for that.

CalMacCQ:
Is there anything we need to add for`OpType.CnRy `?

yao-cqc (approved):
Thanks! I think this is good for now. If you are up for it, it might be worth checking if there are any other unsupported conversions that can be supported.

PR #476, opened 2022-08-24
Add post routing KAK

yao-cqc:


ss2165 (approved):
Should this be added to other backends? Quantinuum?

yao-cqc:
> Should this be added to other backends? Quantinuum?

Sure, I'll go through other backends.

ss2165:
Quantinuum probably isn't a good fit as it doesn't need routing

PR #479, opened 2022-09-06
Support ECRGate in `pytket-qiskit` circuit conversion

burgholzer:
At the moment, attempting to convert a tket circuit containing `ecr` gates (one of the basis gates in OQC's quantum computers) to a qiskit circuit (via the `tk_to_qiskit` method) raises the following error
```console
NotImplementedError: Cannot convert tket Op to Qiskit gate: ECR
```
This simple PR fixes the underlying issue by adding the `ECRGate` to the `_qiskit_gates_2q` list in the qiskit converter.
As a result, circuits compiled to the OQC device using tket can be translated to Qiskit without errors.

### DIFF
--- a/modules/pytket-qiskit/docs/changelog.rst
+++ b/modules/pytket-qiskit/docs/changelog.rst
@@ -5,6 +5,7 @@ Changelog
 -------------------
 
 * Add post-routing ``KAKDecomposition`` to default pass with ``optimisation_level`` = 2.
+* Add support for ``ECRGate`` in ``tk_to_qiskit`` conversion.
 
 0.28.0 (August 2022)
 --------------------
--- a/modules/pytket-qiskit/pytket/extensions/qiskit/qiskit_convert.py
+++ b/modules/pytket-qiskit/pytket/extensions/qiskit/qiskit_convert.py
@@ -119,6 +119,7 @@ _qiskit_gates_2q = {
     qiskit_gates.CXGate: OpType.CX,
     qiskit_gates.CYGate: OpType.CY,
     qiskit_gates.CZGate: OpType.CZ,
+    qiskit_gates.ECRGate: OpType.ECR,
     qiskit_gates.iSwapGate: OpType.ISWAPMax,
     qiskit_gates.RXXGate: OpType.XXPhase,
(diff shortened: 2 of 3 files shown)


ss2165 (changes requested):
Unfortunately, the OQC ECR gate and the Qiskit OCR gate are not exactly the same. [Qiskit defines it as RZX(pi/2)](https://qiskit.org/documentation/stubs/qiskit.circuit.library.ECRGate.html), but the ECR OpType in pytket, which maps to OQC, is RZX(-pi/2)

image
image

https://cqcl.github.io/tket/pytket/api/optype.html

The two gates are not equivalent up to global phase, and extra single qubit gates are needed to map them. (I believe a Z gate on one qubit and an X on the other). You could potentially add some special case handling for this, and use a unitary simulator to verify the conversion.

burgholzer:
Thanks for pointing that out!

Please correct me if I am wrong, but I believe the difference is much simpler than that. The answer actually is in the link to the Qiskit definition you provided above (https://qiskit.org/documentation/stubs/qiskit.circuit.library.ECRGate.html).
Namely, Qiskit's little endian qubit ordering convention makes the difference, i.e., 
```
ECR_QISKIT(q[0], q[1]) = ECR_TKET(q[1], q[0])
```
If that is correct, special casing should be much easier. I'll happily work on this, but would first wait for your confirmation.

ss2165:
Yes you're right, my answer is outdated and we have since updated ECR to be compatible. My colleague @yao-cqc will confirm this and reply.

yao-cqc:
@burgholzer You are correct, the two ECR gates have the same definition. Endianness only affect how we present the matrices. I believe no special case handling is required.

burgholzer:
thanks for the clarification! Then this PR should be complete as it is, right?

yao-cqc:
> thanks for the clarification! Then this PR should be complete as it is, right?

Yes, I think it's good. It would be great if you can also add an entry to `modules/pytket-qiskit/docs/changelog.rst`.

yao-cqc (approved):
All looks good to me. Thanks!

ss2165 (approved):

ss2165:
Thanks @burgholzer!

Issue #480, opened 2022-09-14
pytket-qiskit issue with parameterized circuits


adamcallison:
After installing directly from this repo, I attempted to use `pytket-qiskit` to compile a parameterized `qiskit` circuit. The conversion to `pytket` format via `qiskit_to_tk` worked well, as did the actual compilation passes, but I encountered an error when converting back via `tk_to_qiskit`. As `append_tk_command_to_qiskit` attempts to rebuild the global phase on the circuit, the line
https://github.com/CQCL/pytket-extensions/blob/39a5d17360ad8c11baacaf8f3a4e391d5a02cf49/modules/pytket-qiskit/pytket/extensions/qiskit/qiskit_convert.py#L534
causes `qiskit`'s `parameterexpression.py` to attempt to apply the `+=` operation, and the line
```
elif isinstance(other, numbers.Number) and numpy.isfinite(other):
```
gives the error
```
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```
presumably, because it cannot handle the `sympy` object.

My inelegant workaround is to replace
https://github.com/CQCL/pytket-extensions/blob/39a5d17360ad8c11baacaf8f3a4e391d5a02cf49/modules/pytket-qiskit/pytket/extensions/qiskit/qiskit_convert.py#L534
with
```
    if isinstance(qcirc.global_phase, ParameterExpression):
        qcirc.global_phase += phase*np.pi
    else:
        qcirc.global_phase += phase * sympy.pi
```

adamcallison:
I figured that the `sympy` symbolic `pi` was being used to avoid accumulating floating point errors or similar. I'm happy to do a PR if you're happy with this fix?

cqc-alec:
That's a good point. Could you make a PR, we'll at least see if all the tests pass?
Thanks!

cqc-melf:
See https://github.com/CQCL/pytket-extensions/pull/482

PR #482, opened 2022-09-15
use pi from numpy instead of sympy when rebuilding qiskit global phas…

adamcallison:
…e in pytket-qiskit

As discussed in this issue:
https://github.com/CQCL/pytket-extensions/issues/480


cqc-alec (changes requested):
Thanks. Could you please add a test case that would have failed without this change?

cqc-melf:
I have tried to find a case where we have an error without this fix, but I have not been able to find that. If this is still a problem please reopen this PR at: https://github.com/CQCL/pytket-qiskit/

PR #485, opened 2022-09-26
Add missing backend info for Qulacs / Cirq / PySimplex / Stim backends

isobelhooper:
QulacsBackend and the Cirq backends previously had the `backend_info` property set as `None`, and StimBackend and SimplexBackend did not have the `backend_info` property set at all.

This PR makes the `.backend_info` property of these backends return `BackendInfo` instances that have the name, extension version and gatesets set. (The other properties aren't very useful for locally-run simulators, as far as I can tell, so I've not set them.)

### DIFF
--- a/modules/pytket-qulacs/pytket/extensions/qulacs/backends/qulacs_backend.py
+++ b/modules/pytket-qulacs/pytket/extensions/qulacs/backends/qulacs_backend.py
@@ -123,7 +123,7 @@ class QulacsBackend(Backend):
 
     @property
     def backend_info(self) -> Optional["BackendInfo"]:
-        return None
+        return self._backend_info
 
     @property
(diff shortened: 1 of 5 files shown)


cqc-alec (approved):

modules/pytket-qulacs/pytket/extensions/qulacs/backends/qulacs_backend.py
```diff
@@ -123,7 +123,7 @@ def _result_id_type(self) -> _ResultIdTuple:
 
     @property
     def backend_info(self) -> Optional["BackendInfo"]:
-        return None
+        return self._backend_info
```
cqc-melf, line 126: I am having issues with the manual run, see pytket repo.

PR #494, opened 2022-10-11
clean up, remove modules which are now in new repos

cqc-melf:
will clean up the documentation in another PR, would like to get this merged soon to make sure that no one pushes things here that gets removed and not transferred later.


cqc-melf:
The readme is already updated with links to the new repos.

cqc-alec:
Do you understand the pylint error, is this new?

cqc-melf:
The two `errors` are not new, they are show in every run and not the cause of the problem.
I am talking about this two:
.pylintrc:1:0: R0022: Useless option value for '--enable', 'mixed-indentation' was removed from pylint, see https://github.com/PyCQA/pylint/pull/3578. (useless-option-value)
.pylintrc:1:0: W0012: Unknown option value for '--enable', expected a valid pylint message and got 'relative-import' (unknown-option-value)

See for example here:
https://github.com/CQCL/pytket-extensions/actions/runs/3220581761/jobs/5267462923

cqc-melf:
I would assume that the problem for the stop of the check is that there are no file that can be checked any more in that directory. I have disabled the check and would merge this anyway?

cqc-alec:
> I would assume that the problem for the stop of the check is that there are no file that can be checked any more in that directory. I have disabled the check and would merge this anyway?

OK, maybe we should remove the lint check completely then.

cqc-melf:
> > I would assume that the problem for the stop of the check is that there are no file that can be checked any more in that directory. I have disabled the check and would merge this anyway?
> 
> OK, maybe we should remove the lint check completely then.

I have deleted the file now

cqc-alec (approved):

PR #495, opened 2022-10-12
Fix/docs

cqc-melf:
try to clean up everything in the repo, only keeping the docs and the readme. Is there anything else we should keep at the moment?


cqc-melf:
Any objections on removing the required checks and removing all secrets? @cqc-alec

cqc-alec:
> Any objections on removing the required checks and removing all secrets? @cqc-alec

No objections.

cqc-melf:
> > Any objections on removing the required checks and removing all secrets? @cqc-alec
> 
> No objections.

done

generate_mgit.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
cqc-melf, line 1: not sure if this is something other people would use and not sure if this is the right place to have it.

.github/workflows/docs/build-docs
```diff
@@ -59,16 +71,24 @@ def build_module_docs(module):
         content = f.readlines()
         content.append("\n.. toctree::\n\t:caption: Extensions:\n\t:maxdepth: 1\n\n")
         content.append(f"\tIntroduction <../index.html#http://>\n")
-        for mod in get_all_modules():
```
cqc-alec, line 62: I guess `get_all_modules()` and maybe some other functions can be deleted?

generate_mgit.sh
cqc-alec: Seems as good a place as any, and could be useful to others! But maybe put it inside a directory called e.g. `dev-utils`?

cqc-alec (changes requested):

generate_mgit.sh
cqc-melf: done

.github/workflows/docs/build-docs
cqc-melf: I have done some clean up work in that file, removed this and other things.

cqc-alec (approved):

PR #496, opened 2022-10-12
update main

cqc-melf:
update main from develop


cqc-melf:
I think they are old, but not 100 % sure, will take a closer look. Should I update directly on this branch or on develop?

cqc-alec:
> I think they are old, but not 100 % sure, will take a closer look. Should I update directly on this branch or on develop?

Thanks, Probably the warnings are harmless. May as well be on this branch I guess.

cqc-melf:
I have checked the warnings and renamed the workflow. From my understanding the warnings are nothing to worried about and I would keep this as it is. If you want to discuss this in more details, please let me know.

cqc-alec (approved):

PR #501, opened 2022-10-17
update github actions

cqc-melf:


.github/workflows/docs.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       with:
         fetch-depth: '0'
     - name: Set up Python 3.9
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
```
CalMacCQ, line 17: Can I ask what this change does?

cqc-melf: Updates to the newest version of the action, you can take a look on the changes between them here: https://github.com/actions/setup-python/releases

PR #502, opened 2022-10-21
fix broken extensions links

CalMacCQ:
It was pointed out that many links in the pytket-extensions index page are broken. I missed this before. This should fix the issue.

There is also an issue with the docs building locally. Could there be an issue with the build-docs script? 
Is this correct `DOCS_DIR = Path(sys.argv[0]).absolute().parent` ?

To clarify - if I build the docs locally and try to open the index.html file it doesn't work


cqc-melf:
I am happy to merge this and try if it works? I am not expecting any issues in the docs build from this changes.

cqc-melf (approved):

PR #521, opened 2023-11-24
update links

cqc-melf:


cqc-alec (changes requested):
I think some of these are incorrect because not all extensions exist on the new website. Maybe @CalMacCQ can confirm which ones do and don't.

CalMacCQ:
All of the extensions are built on the tket website with the exception of the following.

* pytket-ionq
* pytket-aqt
* pytket-qujax
* pytket-qsharp

cqc-melf:
I have removed ionq, I will keep the others links as they are, we will move them over as well soon.

PR #522, opened 2023-11-24
Update README.

cqc-alec:
I propose we archive this repo after this. I've closed the remaining issues and would also close #521 before doing so. Any objections?


CalMacCQ:
No objections here.

Previously the tket-site used this repository as source before the extensions index was added to the API docs.

The tket-site no longer uses this repository for source content. It uses the tket repo.

CalMacCQ (approved):

cqc-melf:
Sounds good to me

cqc-melf (approved):

